PR #40, opened 2024-07-08
client support tls connections

famarting:
This PR adds support for tls connections. The change is made in a non breaking manner by adding the parameter as an optional and at the end of the constructors, so existing client instantiations will continue compiling without problems.

The idea for this change is to use it in the Dapr js SDK for workflows... currently the normal Dapr js sdk supports TLS connections but the workflows SDK (based on durabletask) does not support TLS https://github.com/dapr/js-sdk/blob/main/src/workflow/client/DaprWorkflowClient.ts#L48

### DIFF
--- a/src/client/client-grpc.ts
+++ b/src/client/client-grpc.ts
@@ -6,11 +6,13 @@ import * as grpc from "@grpc/grpc-js";
 
 export class GrpcClient {
   private readonly _hostAddress: string;
+  private readonly _tls: boolean;
   private readonly _options: grpc.ChannelOptions;
   private _stub: stubs.TaskHubSidecarServiceClient;
 
-  constructor(hostAddress: string = "localhost:4001", options: grpc.ChannelOptions = {}) {
+  constructor(hostAddress: string = "localhost:4001", options: grpc.ChannelOptions = {}, useTLS: boolean = false) {
     this._hostAddress = hostAddress;
+    this._tls = useTLS;
     this._options = this._generateChannelOptions(options);
     this._stub = this._generateClient();
   }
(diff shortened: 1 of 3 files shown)


cgillum:
The change looks fine to me. Alternatively, could we enable TLS whenever we see that the address is HTTPS (and not HTTP), and not introduce the flag?

famarting:
> The change looks fine to me. Alternatively, could we enable TLS whenever we see that the address is HTTPS (and not HTTP), and not introduce the flag?

I agree, that solution would also be fine. However it would require parsing logic similar to this https://github.com/dapr/go-sdk/blob/main/client/internal/parse.go#L31 in order to support the many ways in which tls can be specified in grpc

cgillum:
It just occurred to me that the endpoint configuration doesn't include the scheme (like http or https), so you can't really use that for a TLS hint. I'm okay with this as a simple enough solution.

cgillum (approved):

cgillum:
Looks like the test and build pipeline is failing. @kaibocai, is this something you're able to take a look at?

kaibocai:
Hi @famarting, thanks for your contribution. Can you try rebasing your PR against the main branch? We had a fix on the CI let's see if that makes the CI green for your PR. Thanks.

kaibocai (approved):

famarting:
updated

TonyPythoneer:
Hi @cgillum,

I fixed that in #42 when there was a timeout issue in the test.
I believe this PR can pass the tests after my PR merges.

---

Hi @famarting,

Amazing! It's passed now! 🥳

---

Hi @kaibocai,

I think this PR can move forward as it gets updated. 😊

famarting:
Hi all! Thank you for merging the PR, my intention was to use this change in the Dapr js sdk to implement a new feature.

Could any of you please create a new release of this library so I can use it from the Dapr js sdk?

Issue #64, opened 2024-10-09
grpc worker stream connection should be retried


famarting:
Equivalent to https://github.com/microsoft/durabletask-go/pull/72

The GRPC stream connection created by the worker client should be infinitely retried. Without this connection orchestrations wont execute, so its key that client retries this connection or at least returns an error to the caller so the worker can be re-started....

PR #66, opened 2024-10-17
retry workflows grpc stream

famarting:
implementation for https://github.com/microsoft/durabletask-js/issues/64

adds infinite retries to the workflows grpc stream connection


kaibocai:
Thanks for the contribution!

PR #68, opened 2024-11-06
support set custom status

famarting:
first step for https://github.com/dapr/js-sdk/issues/638

adds support to set custom status to orchestration context and updates tests


famarting:
ping

PR #75, opened 2026-01-28
Durable Entities Support

YunchuWang:
This pull request introduces comprehensive support and documentation for Durable Entities (stateful actors) in the SDK, including new sample code and expanded README sections. Durable Entities allow users to define, access, and manipulate persistent state objects both from orchestrations and directly from clients, supporting atomic operations and entity locking for critical sections. The changes include new example files, updates to documentation, and minor dependency additions.

**Durable Entities Feature and Documentation:**

* Added detailed Durable Entities documentation to `README.md`, including how to define entities, access them from clients and orchestrations, perform entity locking for atomic operations, and clean up entities. Also added usage examples and links to new sample files. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R96-R123) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R152-R155) [[3]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R164-R277)
* Added a new section to the hello-world examples README describing the new entity samples: `entity-counter.ts` and `entity-orchestration.ts`.

**New Durable Entities Sample Code:**

* Added `examples/hello-world/entity-counter.ts`, demonstrating how to define a simple counter entity, perform operations, and interact with it from a client.
* Added `examples/hello-world/entity-orchestration.ts`, showing how to use entities within orchestrations, including entity locking for atomic bank transfers and signaling entities.

**SDK and Dependency Updates:**

* Updated the client import in `packages/durabletask-js/src/client/client.ts` to include `Int32Value` from the protobuf wrappers, likely in support of new entity features.

**Other Minor Changes:**

* Removed a duplicate `category` line from the CodeQL workflow YAML.

### DIFF
--- a/packages/durabletask-js/test/entity-client.spec.ts
+++ b/packages/durabletask-js/test/entity-client.spec.ts
@@ -286,6 +286,43 @@ describe("Entity Client Proto Conversion", () => {
       expect(metadata.getLockedby()).toBeUndefined();
       expect(metadata.getSerializedstate()).toBeUndefined();
     });
+
+    it("should gracefully handle invalid JSON in serialized state", () => {
+      // Arrange - simulate what convertEntityMetadata does
+      const metadata = new pb.EntityMetadata();
+      metadata.setInstanceid("@counter@test");
+      metadata.setBacklogqueuesize(0);
+
+      const ts = new Timestamp();
+      ts.fromDate(new Date());
+      metadata.setLastmodifiedtime(ts);
+
+      // Set invalid JSON as state
+      const state = new StringValue();
+      state.setValue("{invalid json}");
+      metadata.setSerializedstate(state);
+
+      const serializedState = metadata.getSerializedstate()?.getValue();
+      const includeState = true;
+
+      // Act - replicate the same logic as convertEntityMetadata
+      let parsedState: unknown = undefined;
+      let parseSucceeded = false;
+
+      if (includeState && serializedState) {
+        try {
+          parsedState = JSON.parse(serializedState);
+          parseSucceeded = true;
+        } catch {
+          // Should fall through gracefully
+          parseSucceeded = false;
+        }
+      }
+
+      // Assert - invalid JSON should not throw, should fallback
+      expect(parseSucceeded).toBe(false);
+      expect(parsedState).toBeUndefined();
+    });
   });
 });
 
--- a/packages/durabletask-js/test/entity-executor.spec.ts
+++ b/packages/durabletask-js/test/entity-executor.spec.ts
@@ -398,4 +398,104 @@ describe("TaskEntityShim", () => {
       expect(failure.getEndtimeutc()).toBeDefined();
     });
   });
+
+  describe("StateShim error handling", () => {
+    it("should throw when setting non-serializable state", async () => {
+      // Entity that sets state to a circular reference
+      class CircularStateEntity extends TaskEntity<unknown> {
+        makeCircular(): void {
+          const obj: Record<string, unknown> = {};
+          obj.self = obj; // Circular reference
+          this.state = obj;
+        }
+
+        protected initializeState(): unknown {
+          return {};
+        }
+      }
+
+      const entity = new CircularStateEntity();
+      const shim = new TaskEntityShim(entity, entityId);
+      const request = createBatchRequest(
+        entityId.toString(),
+        [{ name: "makeCircular" }],
+        {},
+      );
+
+      const result = await shim.executeAsync(request);
+
+      // The operation should fail because setState will throw on JSON.stringify
+      const opResult = result.getResultsList()[0];
+      expect(opResult.hasFailure()).toBe(true);
+
+      const failure = opResult.getFailure()!;
+      expect(failure.getFailuredetails()?.getErrormessage()).toContain(
+        "Entity state is not JSON-serializable",
+      );
+    });
+  });
+
+  describe("invalid operation input", () => {
+    it("should record failure when operation input is invalid JSON", async () => {
+      const entity = new CounterEntity();
+      const shim = new TaskEntityShim(entity, entityId);
+
+      // Manually create a batch request with invalid JSON input
+      const request = new pb.EntityBatchRequest();
+      request.setInstanceid(entityId.toString());
+
+      const stateValue = new StringValue();
+      stateValue.setValue(JSON.stringify({ count: 0 }));
+      request.setEntitystate(stateValue);
+
+      const opRequest = new pb.OperationRequest();
+      opRequest.setOperation("add");
+      const inputValue = new StringValue();
+      inputValue.setValue("{invalid json"); // Invalid JSON
+      opRequest.setInput(inputValue);
+      request.addOperations(opRequest);
+
+      const result = await shim.executeAsync(request);
+
+      const opResult = result.getResultsList()[0];
+      expect(opResult.hasFailure()).toBe(true);
+    });
+
+    it("should continue executing after invalid JSON input failure", async () => {
+      const entity = new CounterEntity();
+      const shim = new TaskEntityShim(entity, entityId);
+
+      // Create request with invalid JSON for first op, then a valid op
+      const request = new pb.EntityBatchRequest();
+      request.setInstanceid(entityId.toString());
+
+      const stateValue = new StringValue();
+      stateValue.setValue(JSON.stringify({ count: 10 }));
+      request.setEntitystate(stateValue);
+
+      // First op: invalid JSON input
+      const op1 = new pb.OperationRequest();
+      op1.setOperation("add");
+      const badInput = new StringValue();
+      badInput.setValue("not valid json!");
+      op1.setInput(badInput);
+      request.addOperations(op1);
+
+      // Second op: valid operation
+      const op2 = new pb.OperationRequest();
+      op2.setOperation("get");
+      request.addOperations(op2);
+
+      const result = await shim.executeAsync(request);
+
+      expect(result.getResultsList()).toHaveLength(2);
+
+      // First operation fails due to invalid JSON
+      expect(result.getResultsList()[0].hasFailure()).toBe(true);
+
+      // Second operation succeeds, state should be rolled back to 10
+      expect(result.getResultsList()[1].hasSuccess()).toBe(true);
+      expect(JSON.parse(result.getResultsList()[1].getSuccess()!.getResult()!.getValue())).toBe(10);
+    });
+  });
 });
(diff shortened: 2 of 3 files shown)


packages/durabletask-js/src/worker/entity-executor.ts
```diff
@@ -0,0 +1,471 @@
...
+
+  /**
+   * Executes a single operation with transactional semantics.
+   */
+  private async executeOperation(opRequest: pb.OperationRequest): Promise<pb.OperationResult> {
+    const startTime = new Date();
+
+    // Parse operation input
+    const inputValue = opRequest.getInput();
+    const input = inputValue ? JSON.parse(inputValue.getValue()) : undefined;
```
torosent: 🔴 **Bug: `JSON.parse` outside per-operation try-catch — fails entire batch on malformed input**

This `JSON.parse` is **outside** the per-operation try-catch block (which starts at line 352). If a single operation has malformed JSON input, the exception propagates up to `executeAsync` (no catch there), then to `_executeEntity` which catches it as a framework-level error — **failing the entire batch** instead of just the one bad operation.

This breaks the transactional contract documented in the class: "Each operation is executed independently" and "Other operations in the batch continue to execute."

**Suggested fix:** Move the JSON.parse inside the try-catch block:
```typescript
try {
  const inputValue = opRequest.getInput();
  const input = inputValue ? JSON.parse(inputValue.getValue()) : undefined;
  this.operationShim.setNameAndInput(opRequest.getOperation(), input);
  const output = await Promise.resolve(this.entity.run(this.operationShim));
  // ...
```

packages/durabletask-js/src/worker/entity-executor.ts
```diff
@@ -0,0 +1,471 @@
...
+
+      // Create failure result
+      const failure = new pb.OperationResultFailure();
+      const failureDetails = new pb.TaskFailureDetails();
+
+      if (error instanceof Error) {
+        failureDetails.setErrortype(error.name);
+        failureDetails.setErrormessage(error.message);
+        if (error.stack) {
+          failureDetails.setStacktrace(new StringValue().setValue(error.stack));
```
torosent: 🟡 **Bug: Stack trace may be lost — relies on undocumented method chaining**

`new StringValue().setValue(error.stack)` chains the `setValue()` call, but `google-protobuf`'s TypeScript declarations type `setValue()` as returning `void`. This relies on the undocumented runtime behavior that the JS implementation returns `this`.

Compare to the correct two-step pattern used elsewhere in this same file (lines 326-327):
```typescript
const stateValue = new StringValue();
stateValue.setValue(finalState);
batchResult.setEntitystate(stateValue);
```

**Suggested fix:**
```typescript
if (error.stack) {
  const stackValue = new StringValue();
  stackValue.setValue(error.stack);
  failureDetails.setStacktrace(stackValue);
}
```

packages/durabletask-js/src/entities/task-entity.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Finds a method on this class that matches the operation name (case-insensitive).
+   *
+   * @param operationName - The operation name (already lowercased).
+   * @returns The actual method name if found, undefined otherwise.
+   */
+  private findMethod(operationName: string): string | undefined {
+    // Get all own property names of this instance and its prototype chain
+    const proto = Object.getPrototypeOf(this);
```
torosent: 🔴 **Bug: `findMethod` only searches immediate prototype — breaks multi-level inheritance**

`Object.getOwnPropertyNames(proto)` only returns own properties of the **direct** prototype. If a user creates a multi-level hierarchy like `CounterEntity extends BaseEntity extends TaskEntity`, methods defined on `BaseEntity` won't be found, and operations will throw `"No suitable method found"`.

**Suggested fix:** Walk the prototype chain in a loop:
```typescript
private findMethod(operationName: string): string | undefined {
  let proto = Object.getPrototypeOf(this);
  const taskEntityProto = TaskEntity.prototype;
  while (proto && proto !== taskEntityProto) {
    const methodNames = Object.getOwnPropertyNames(proto);
    for (const name of methodNames) {
      if (name.toLowerCase() === operationName) {
        const prop = (this as unknown as Record<string, unknown>)[name];
        if (typeof prop === "function" && name !== "constructor" && name !== "run") {
          return name;
        }
      }
    }
    proto = Object.getPrototypeOf(proto);
  }
  return undefined;
}
```

packages/durabletask-js/src/entities/task-entity.ts
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * Initializes the entity state. This is only called when there is no current state.
+   *
+   * @returns The initial entity state.
+   *
+   * @remarks
+   * The default implementation returns an empty object cast to TState.
+   * Override this method to provide custom initialization logic.
+   */
+  protected initializeState(): TState {
```
torosent: 🟡 **Correctness: `initializeState()` returns `{} as TState` — silent type violations**

If `TState` is `number`, `string`, an array, or an interface with required properties, the default empty object `{}` will silently violate the type contract at runtime. For example, `TaskEntity<number>` would get `{}` instead of a number.

Consider making `initializeState()` abstract to force subclasses to provide a correct initial state, or at minimum document this limitation prominently.

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
```diff
@@ -571,3 +585,373 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
...
+
+    // Initialize critical section state (availableEntities is empty until lock is granted)
+    this.criticalSection = {
+      id: criticalSectionId,
+      lockedEntities: uniqueEntities,
+      availableEntities: new Set<string>(), // Empty until lock is granted
+    };
+
+    // Mark that we're waiting for the lock to be granted
+    this.lockAcquisitionPending = true;
```
torosent: 🟡 **Robustness: `lockAcquisitionPending` flag never reset on failure/termination paths**

`lockAcquisitionPending` is set to `true` here but only cleared in `completeLockAcquisition()` (line 870). If:
- The orchestration is terminated before the lock is granted
- The lock acquisition fails for any reason
- The orchestration is replayed in a state where the grant event is missing

…the flag stays `true` forever, blocking all subsequent entity calls with: *"Must await the completion of the lock request prior to calling any entity."*

There's no `ENTITYLOCKFAILED` event handler in `orchestration-executor.ts`.

Consider resetting the flag in `exitCriticalSection()` and/or adding a timeout/failure handler for lock acquisition.

packages/durabletask-js/src/worker/entity-executor.ts
```diff
@@ -0,0 +1,471 @@
...
+        this.serializedValue !== undefined ? JSON.parse(this.serializedValue) : undefined;
+      this.cacheValid = true;
+    }
+
+    return this.cachedValue as T;
+  }
+
+  setState(state: unknown): void {
+    this.cachedValue = state;
+    this.serializedValue = state !== undefined && state !== null ? JSON.stringify(state) : undefined;
```
torosent: 🟡 **Robustness: `setState` silently fails on non-serializable values**

`JSON.stringify(state)` will throw on circular references and silently drop functions/Symbols/`undefined` values within objects. Since `setState` is called from user entity code, the error message from a raw `TypeError: Converting circular structure to JSON` won't help users debug the issue.

Consider wrapping this in a try-catch with a descriptive error:
```typescript
setState(state: unknown): void {
  this.cachedValue = state;
  try {
    this.serializedValue = state !== undefined && state !== null ? JSON.stringify(state) : undefined;
  } catch (e) {
    throw new Error(`Entity state is not JSON-serializable: ${e instanceof Error ? e.message : String(e)}`);
  }
  this.cacheValid = true;
}
```

packages/durabletask-js/src/client/client.ts
```diff
@@ -964,6 +974,250 @@ export class TaskHubGrpcClient {
...
+    const instanceIdStr = protoMetadata.getInstanceid();
+    const entityId = EntityInstanceId.fromString(instanceIdStr);
+
+    const lastModifiedTime = protoMetadata.getLastmodifiedtime()?.toDate() ?? new Date();
+    const backlogQueueSize = protoMetadata.getBacklogqueuesize();
+    const lockedBy = protoMetadata.getLockedby()?.getValue();
+    const serializedState = protoMetadata.getSerializedstate()?.getValue();
+
+    if (includeState && serializedState) {
+      const state = JSON.parse(serializedState) as T;
```
torosent: 🔴 **Bug: Unprotected `JSON.parse` — corrupted server state crashes client**

If the server returns corrupted or invalid JSON in the entity state, this `JSON.parse` will throw an unhandled error that crashes the entire `getEntity()` call or breaks the `getEntities()` pagination iterator.

**Suggested fix:**
```typescript
if (includeState && serializedState) {
  try {
    const state = JSON.parse(serializedState) as T;
    return createEntityMetadata<T>(entityId, lastModifiedTime, backlogQueueSize, lockedBy, state);
  } catch {
    // Return metadata without state if parsing fails
    return createEntityMetadataWithoutState(entityId, lastModifiedTime, backlogQueueSize, lockedBy) as EntityMetadata<T>;
  }
}
```

Or throw a more descriptive error wrapping the parse failure.

packages/durabletask-js/src/worker/entity-executor.ts
YunchuWang: done

packages/durabletask-js/src/client/client.ts
YunchuWang: done

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
YunchuWang: This is local in-memory state that only exists for the lifetime of a single orchestration replay. The .NET implementation (OrchestrationEntityContext.cs) has the exact same pattern — lockAcquisitionPending is set in EmitAcquireMessage() and only cleared in CompleteAcquire().

Orchestration terminated: The entire RuntimeOrchestrationContext is discarded — no state persists.
Lock acquisition fails: There's no ENTITYLOCKFAILED event in the protocol (same as .NET). The orchestration either times out or gets terminated, both of which destroy the context.
Replay with missing grant event: The orchestration yields at the await lockEntities() call and doesn't continue — the flag being true is correct because the lock genuinely hasn't been granted yet.

PR #96, opened 2026-02-06
Retry Handler Support when call activity or suborchestration

YunchuWang:
# Summary

### New Features: Retry Handler Support

This PR adds support for imperative retry control through `AsyncRetryHandler` and `RetryHandler` functions, allowing users to make custom retry decisions based on failure context. This mirrors the .NET SDK's `InvokeWithCustomRetryHandler` pattern.

**New Types & APIs:**
- `AsyncRetryHandler` - Async function delegate for imperative retry decisions
- `RetryHandler` - Sync function delegate for imperative retry decisions  
- `RetryContext` - Context object passed to handlers with failure details, attempt count, elapsed time
- `toAsyncRetryHandler()` - Utility to normalize sync handlers to async
- `taskOptionsFromRetryHandler()` - Creates TaskOptions from an AsyncRetryHandler
- `taskOptionsFromSyncRetryHandler()` - Creates TaskOptions from a sync RetryHandler
- `TaskRetryOptions` - Union type supporting `RetryPolicy | AsyncRetryHandler | RetryHandler`

**Enhanced RetryPolicy:**
- Added `handleFailure` predicate option for filtering retryable failures by error type/message
- Added `retryTimeoutInMilliseconds` option for overall retry timeout
- Added `shouldRetry()` method to evaluate failures against the predicate

**New Internal Classes:**
- `RetryTaskBase<T>` - Abstract base class for retry tasks, shared between policy and handler approaches
- `RetryHandlerTask<T>` - Task implementation for handler-based retry (immediate reschedule, no timer delay)

**Orchestration Context Changes:**
- `callActivity()` and `callSubOrchestrator()` now accept `TaskOptions.retry` as `RetryPolicy`, `AsyncRetryHandler`, or `RetryHandler`
- Added `rescheduleRetryTask()` for immediate retry rescheduling (used by handler-based retry)

### Bug Fixes
- Fixed error type detection to use `e.name` instead of `e.constructor.name`, preserving custom error names set via `error.name = "CustomType"`
- Fixed retry logging to only log "Retrying task" after the handler approves the retry

## Why is this change needed?

1. **Imperative retry control**: Users need fine-grained control over retry decisions that can't be expressed declaratively (e.g., retry based on error message content, external state, or complex conditions)
2. **Error type filtering**: The `handleFailure` predicate enables fail-fast behavior for non-retryable errors (e.g., validation errors, authorization failures)
3. **Parity with .NET SDK**: Aligns with the .NET Durable Functions SDK's `InvokeWithCustomRetryHandler` pattern
4. **Custom error names**: Users can now throw errors with `error.name = "PermanentError"` and filter on that in retry handlers

## Issues / work items
- Resolves #96

---

# Project checklist
- [ ] Release notes are not required for the next release
  - [x] Otherwise: Notes added to `CHANGELOG.md`
- [x] Backport is not required
  - [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
- [x] All required tests have been added/updated (unit tests, E2E tests)
- [ ] Breaking change?
  - [ ] If yes:
    - Impact: None - All changes are additive. Existing `RetryPolicy` usage continues to work unchanged.
    - Migration guidance: N/A

---

# AI-assisted code disclosure (required)

## Was an AI tool used? (select one)
- [ ] No
- [x] Yes, AI helped write parts of this PR (e.g., GitHub Copilot)
- [ ] Yes, an AI agent generated most of this PR

If AI was used:
- Tool(s): GitHub Copilot
- AI-assisted areas/files: 
  - Bug fix in `pb-helper.util.ts` (error name detection)
  - Bug fix in `orchestration-executor.ts` (retry logging order)
- What you changed after AI output: Reviewed and verified fixes address all failing tests

AI verification (required if AI was used):
- [x] I understand the code and can explain it
- [x] I verified referenced APIs/types exist and are correct
- [x] I reviewed edge cases/failure paths (timeouts, retries, cancellation, exceptions)
- [x] I reviewed concurrency/async behavior
- [x] I checked for unintended breaking or behavior changes

---

# Testing

## Automated tests
- Result: **Passed**

### New Unit Tests (7 test files, 100+ test cases):
- `retry-context.spec.ts` - RetryContext creation and properties
- `retry-handler.spec.ts` - RetryHandler/AsyncRetryHandler types and toAsyncRetryHandler()
- `retry-handler-task.spec.ts` - RetryHandlerTask shouldRetry() behavior
- `retry-policy-handle-failure.spec.ts` - handleFailure predicate functionality
- `retryable-task.spec.ts` - RetryableTask with enhanced policy features
- `task-options-retry-handler.spec.ts` - TaskOptions creation helpers
- `orchestration_executor.spec.ts` - Updated for new retry scenarios

### New E2E Tests (2 test files, 15 test cases):
- `retry-handler.spec.ts` - AsyncRetryHandler with activities and sub-orchestrations
- `retry-advanced.spec.ts` - handleFailure predicate, timeout, and combined features

## Manual validation (only if runtime/behavior changed)
- Environment (OS, Node.js version, components): Windows, Node.js 22.x, DTS emulator (Docker)
- Steps + observed results:
  1. `npm run build` - Successful
  2. `npm run lint` - Passed
  3. `npm run test` - All unit tests passed
  4. `npx jest test/e2e-azuremanaged/retry-*.spec.ts --runInBand --forceExit` - All 15 E2E tests passed
- Evidence: Error types like `TransientError`, `PermanentError`, `ValidationError` correctly captured and used in retry decisions

---

# Notes for reviewers

## Key Design Decisions

1. **RetryTaskBase abstraction**: Created a shared base class to avoid code duplication between `RetryableTask` (policy-based) and `RetryHandlerTask` (handler-based)

2. **Immediate vs delayed retry**: 
   - `RetryableTask` (policy) uses timers for backoff delays
   - `RetryHandlerTask` (handler) reschedules immediately - delay logic is the handler's responsibility

3. **Handler runs as orchestrator code**: Like .NET SDK, the retry handler is subject to replay determinism requirements. `RetryContext.orchestrationContext.isReplaying` is exposed for guarding side-effects.

4. **Sync/Async handler support**: Both sync `RetryHandler` and async `AsyncRetryHandler` are supported. Sync handlers are automatically wrapped via `toAsyncRetryHandler()`.

## Files Changed Summary

| Category | Files |
|----------|-------|
| New types/APIs | `retry-handler.ts`, `retry-context.ts`, `task-options.ts` |
| Enhanced policy | `retry-policy.ts`, `retryable-task.ts` |
| New task class | `retry-handler-task.ts`, `retry-task-base.ts` |
| Executor integration | `orchestration-executor.ts`, `runtime-orchestration-context.ts` |
| Bug fixes | `pb-helper.util.ts` |
| Exports | `index.ts`, `retry/index.ts`, `options/index.ts` |
| Unit tests | 7 new/updated test files |
| E2E tests | 2 new test files |

### DIFF
--- a/packages/durabletask-js/src/task/retryable-task.ts
+++ b/packages/durabletask-js/src/task/retryable-task.ts
@@ -59,6 +59,11 @@ export class RetryableTask<T> extends RetryTaskBase<T> {
    * The delay is capped at maxRetryInterval.
    */
   computeNextDelayInMilliseconds(currentTime: Date): number | undefined {
+    // Check for non-retriable failures first (e.g., activity not found, version mismatch)
+    if (this.lastFailure?.getIsnonretriable()) {
+      return undefined;
+    }
+
     // Check if handleFailure predicate says we should NOT retry this failure type
     if (this.lastFailure) {
       const failureDetails = {
--- a/packages/durabletask-js/src/worker/orchestration-executor.ts
+++ b/packages/durabletask-js/src/worker/orchestration-executor.ts
@@ -588,8 +588,13 @@ export class OrchestrationExecutor {
         task.incrementAttemptCount();
 
         if (typeof retryResult === "number") {
-          // Handler returned a delay in milliseconds — use a timer
-          ctx.createRetryTimer(task, retryResult);
+          if (retryResult <= 0) {
+            // Handler returned true — retry immediately
+            ctx.rescheduleRetryTask(task);
+          } else {
+            // Handler returned a delay in milliseconds — use a timer
+            ctx.createRetryTimer(task, retryResult);
+          }
         } else {
           // Handler returned true — retry immediately
           ctx.rescheduleRetryTask(task);
--- a/packages/durabletask-js/src/worker/runtime-orchestration-context.ts
+++ b/packages/durabletask-js/src/worker/runtime-orchestration-context.ts
@@ -577,7 +577,8 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
       const name = scheduleTask.getName();
       const input = scheduleTask.getInput()?.getValue();
       const tags = mapToRecord(scheduleTask.getTagsMap());
-      newAction = ph.newScheduleTaskAction(newId, name, input, tags);
+      const version = scheduleTask.getVersion()?.getValue();
+      newAction = ph.newScheduleTaskAction(newId, name, input, tags, version);
     } else {
       // Reschedule a sub-orchestration task
       const subOrch = originalAction.getCreatesuborchestration();
@@ -588,7 +589,8 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
       const instanceId = subOrch.getInstanceid();
       const input = subOrch.getInput()?.getValue();
       const tags = mapToRecord(subOrch.getTagsMap());
-      newAction = ph.newCreateSubOrchestrationAction(newId, name, instanceId, input, tags);
+      const version = subOrch.getVersion()?.getValue();
+      newAction = ph.newCreateSubOrchestrationAction(newId, name, instanceId, input, tags, version);
     }
 
     // Register the new action
--- a/test/e2e-azuremanaged/retry-handler.spec.ts
+++ b/test/e2e-azuremanaged/retry-handler.spec.ts
@@ -299,6 +299,238 @@ describe("Retry Handler E2E Tests", () => {
     }, 31000);
   });
 
+  // ==================== Retry Handler with Delay (number return) Tests ====================
+
+  describe("retry handler returning delay in milliseconds", () => {
+    it("should retry activity after specified delay when handler returns a positive number", async () => {
+      let attemptCount = 0;
+      const attemptTimes: number[] = [];
+
+      const flakyActivity = async (_: ActivityContext, input: number) => {
+        attemptCount++;
+        attemptTimes.push(Date.now());
+        if (attemptCount < 3) {
+          throw new Error(`Transient failure on attempt ${attemptCount}`);
+        }
+        return input * 2;
+      };
+
+      // Retry handler returning delay in ms (fixed 500ms delay)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const retryHandler: any = async (ctx: RetryContext): Promise<boolean | number> => {
+        if (ctx.lastAttemptNumber >= 5) {
+          return false;
+        }
+        return 500; // retry after 500ms
+      };
+
+      const orchestrator: TOrchestrator = async function* (ctx: OrchestrationContext, input: number): any {
+        const result = yield ctx.callActivity(flakyActivity, input, { retry: retryHandler });
+        return result;
+      };
+
+      taskHubWorker.addActivity(flakyActivity);
+      taskHubWorker.addOrchestrator(orchestrator);
+      await taskHubWorker.start();
+
+      const id = await taskHubClient.scheduleNewOrchestration(orchestrator, 21);
+      const state = await taskHubClient.waitForOrchestrationCompletion(id, undefined, 30);
+
+      expect(state).toBeDefined();
+      expect(state?.runtimeStatus).toEqual(OrchestrationStatus.ORCHESTRATION_STATUS_COMPLETED);
+      expect(state?.failureDetails).toBeUndefined();
+      expect(state?.serializedOutput).toEqual(JSON.stringify(42));
+      expect(attemptCount).toBe(3);
+
+      // Verify delays are approximately 500ms
+      if (attemptTimes.length >= 3) {
+        const delay1 = attemptTimes[1] - attemptTimes[0];
+        const delay2 = attemptTimes[2] - attemptTimes[1];
+        // Allow generous tolerance for timer scheduling overhead
+        expect(delay1).toBeGreaterThanOrEqual(400);
+        expect(delay2).toBeGreaterThanOrEqual(400);
+      }
+    }, 31000);
+
+    it("should support exponential backoff via handler returning increasing delays", async () => {
+      let attemptCount = 0;
+      const attemptTimes: number[] = [];
+
+      const flakyActivity = async (_: ActivityContext) => {
+        attemptCount++;
+        attemptTimes.push(Date.now());
+        if (attemptCount < 4) {
+          throw new Error(`Failure on attempt ${attemptCount}`);
+        }
+        return "success";
+      };
+
+      // Retry handler implementing manual exponential backoff: 200ms, 400ms, 800ms
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const retryHandler: any = async (ctx: RetryContext): Promise<boolean | number> => {
+        if (ctx.lastAttemptNumber >= 5) {
+          return false;
+        }
+        return 200 * Math.pow(2, ctx.lastAttemptNumber - 1);
+      };
+
+      const orchestrator: TOrchestrator = async function* (ctx: OrchestrationContext): any {
+        const result = yield ctx.callActivity(flakyActivity, undefined, { retry: retryHandler });
+        return result;
+      };
+
+      taskHubWorker.addActivity(flakyActivity);
+      taskHubWorker.addOrchestrator(orchestrator);
+      await taskHubWorker.start();
+
+      const id = await taskHubClient.scheduleNewOrchestration(orchestrator);
+      const state = await taskHubClient.waitForOrchestrationCompletion(id, undefined, 30);
+
+      expect(state).toBeDefined();
+      expect(state?.runtimeStatus).toEqual(OrchestrationStatus.ORCHESTRATION_STATUS_COMPLETED);
+      expect(state?.failureDetails).toBeUndefined();
+      expect(state?.serializedOutput).toEqual(JSON.stringify("success"));
+      expect(attemptCount).toBe(4);
+
+      // Verify delays are increasing
+      if (attemptTimes.length >= 4) {
+        const delay1 = attemptTimes[1] - attemptTimes[0];
+        const delay2 = attemptTimes[2] - attemptTimes[1];
+        const delay3 = attemptTimes[3] - attemptTimes[2];
+
+        // Each delay should be roughly double the previous (with tolerance)
+        expect(delay2).toBeGreaterThan(delay1 * 0.8);
+        expect(delay3).toBeGreaterThan(delay2 * 0.8);
+      }
+    }, 31000);
+
+    it("should retry sub-orchestration with delay when handler returns a number", async () => {
+      let attemptCount = 0;
+
+      // eslint-disable-next-line require-yield
+      const failingSubOrch: TOrchestrator = async function* (_ctx: OrchestrationContext): any {
+        attemptCount++;
+        if (attemptCount < 3) {
+          throw new Error(`Sub-orch failure on attempt ${attemptCount}`);
+        }
+        return "sub-orch-result";
+      };
+
+      // Retry handler returning fixed delay
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const retryHandler: any = async (ctx: RetryContext): Promise<boolean | number> => {
+        if (ctx.lastAttemptNumber >= 5) {
+          return false;
+        }
+        return 300; // 300ms delay
+      };
+
+      const parentOrchestrator: TOrchestrator = async function* (ctx: OrchestrationContext): any {
+        const result = yield ctx.callSubOrchestrator(failingSubOrch, undefined, {
+          retry: retryHandler,
+        });
+        return result;
+      };
+
+      taskHubWorker.addOrchestrator(failingSubOrch);
+      taskHubWorker.addOrchestrator(parentOrchestrator);
+      await taskHubWorker.start();
+
+      const id = await taskHubClient.scheduleNewOrchestration(parentOrchestrator);
+      const state = await taskHubClient.waitForOrchestrationCompletion(id, undefined, 30);
+
+      expect(state).toBeDefined();
+      expect(state?.runtimeStatus).toEqual(OrchestrationStatus.ORCHESTRATION_STATUS_COMPLETED);
+      expect(state?.failureDetails).toBeUndefined();
+      expect(state?.serializedOutput).toEqual(JSON.stringify("sub-orch-result"));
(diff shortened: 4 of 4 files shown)


packages/durabletask-js/src/worker/orchestration-executor.ts
```diff
@@ -564,4 +548,51 @@ export class OrchestrationExecutor {
...
+        ctx.createRetryTimer(task, nextDelayMs);
+        delete ctx._pendingTasks[taskId];
+        return true;
+      }
+    }
+
+    if (task instanceof RetryHandlerTask) {
+      task.recordFailure(errorMessage, failureDetails);
+      const keepRetrying = await task.shouldRetry(ctx._currentUtcDatetime);
+
```
torosent: **Bug (Medium): Sequential `if` should be `else if`**

Both `instanceof` checks run independently. When a `RetryableTask` exhausts its retries (`computeNextDelayInMilliseconds` returns `undefined`), execution falls through into the `RetryHandlerTask` block. Since `RetryableTask` is not a `RetryHandlerTask`, the check is a no-op, but:

1. It's an unnecessary `instanceof` check on every exhausted `RetryableTask`.
2. If the class hierarchy ever changes, this could silently double-process.
3. Makes the intent unclear to readers.

```suggestion
    if (task instanceof RetryHandlerTask) {
```
should be `else if (task instanceof RetryHandlerTask) {`

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
```diff
@@ -567,7 +590,7 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
     // Register the new action
```
torosent: **Correctness (Medium): Stale `action` reference after rescheduling**

`RetryTaskBase._action` is `readonly` and set once in the constructor. When `rescheduleRetryTask()` creates a new action with a new sequence ID, the task still holds the original action reference via `retryTask.action`.

If anything downstream reads `retryTask.action.getId()` after rescheduling, it gets the stale original ID. Currently `retryTask.action` is only used to read name/input (which don't change), so this *works*, but it's fragile. If the `action` getter is used elsewhere in the future, it could silently return outdated data.

Consider either:
1. Making `_action` mutable and updating it after rescheduling, or
2. Adding a JSDoc note on `RetryTaskBase.action` stating it refers to the *original* action and may not reflect the current sequence ID.

packages/durabletask-js/src/utils/pb-helper.util.ts
```diff
@@ -183,8 +187,9 @@ export function newSubOrchestrationFailedEvent(eventId: number, ex: Error): pb.H
 
 export function newFailureDetails(e: unknown): pb.TaskFailureDetails {
   const failure = new pb.TaskFailureDetails();
-
-  const errorType = e instanceof Error ? e.constructor.name : "UnknownError";
+  // Use e.name (which can be customized) over constructor.name (which is always the class name)
```
torosent: **Correctness (High): `e.name` vs `e.constructor.name` is a potential breaking change**

This change from `e.constructor.name` to `e.name` affects how error types are serialized into protobuf, which in turn affects `handleFailure` predicates and retry handler logic.

For custom error classes that **don't** set `this.name` in their constructor:

```typescript
class MyCustomError extends Error {}
// e.name → "Error" (inherited from Error.prototype)
// e.constructor.name → "MyCustomError"
```

After this change, such errors will serialize as `"Error"` instead of `"MyCustomError"`, silently breaking any `handleFailure` predicates that filter by error type.

The PR correctly adds `this.name = "TaskFailedError"` to `TaskFailedError`, but **other Error subclasses in this codebase** are not updated:
- `OrchestrationStateError`
- `OrchestratorNotRegisteredError`
- `StopIterationError`

**Recommendations:**
1. Audit all `Error` subclasses in the repo and add `this.name = "ClassName"` to each.
2. Document this as a **breaking change** in the CHANGELOG.
3. Consider a migration note for users with existing custom error classes.

packages/durabletask-js/src/worker/orchestration-executor.ts
```diff
@@ -564,4 +548,51 @@ export class OrchestrationExecutor {
...
+      if (keepRetrying) {
+        WorkerLogs.retryingTask(this._logger, ctx._instanceId, task.taskName, task.attemptCount);
```
torosent: **Design (High): `RetryHandlerTask` retries are always immediate — no backoff mechanism**

Unlike `RetryableTask` which uses `createRetryTimer` to add a delay between retries, `RetryHandlerTask` is rescheduled **immediately** via `rescheduleRetryTask`. There is no mechanism for the handler to introduce a delay:

- The handler signature `(RetryContext) => Promise<boolean>` only returns a boolean.
- While `RetryContext.orchestrationContext` exposes `createTimer()`, the handler can't yield/await on a timer since it's not a generator.

This means handler-based retries will **spam the backend** with rapid re-executions for transient failures. In contrast, the .NET SDK's equivalent allows awaiting timers within the handler (since it runs as orchestrator code in a generator context).

Consider one of:
1. **Add an optional delay return**: `Promise<boolean | number>` where a number means "retry after N ms"
2. **Add a `getRetryDelay(context): number` callback** alongside the handler
3. **Document this limitation prominently** and recommend users combine `RetryHandler` with `RetryPolicy`-style backoff patterns
4. **Schedule a default minimum delay** (e.g., 1s) between handler-based retries to prevent tight retry loops

packages/durabletask-js/src/task/options/task-options.ts
```diff
@@ -81,6 +96,48 @@ export function taskOptionsFromRetryPolicy(policy: RetryPolicy): TaskOptions {
...
+ *   return context.lastAttemptNumber < 3;
+ * };
+ *
+ * const options = taskOptionsFromSyncRetryHandler(handler);
+ * await ctx.callActivity("myActivity", input, options);
+ * ```
+ */
+export function taskOptionsFromSyncRetryHandler(handler: RetryHandler): TaskOptions {
+  return { retry: toAsyncRetryHandler(handler) };
+}
```
torosent: **Performance (Low): `toAsyncRetryHandler` double-wrapping**

When `taskOptionsFromSyncRetryHandler` is used:
1. It wraps the sync handler via `toAsyncRetryHandler(handler)` **here**.
2. `createRetryTaskOrDefault` wraps again via `toAsyncRetryHandler(options.retry)`.

The double-wrapping is functionally correct (since `Promise.resolve(promise)` returns the same promise), but creates an extra closure allocation per task. 

Consider either:
- Having `createRetryTaskOrDefault` handle normalization exclusively (remove pre-wrapping here), or
- Storing the handler directly: `return { retry: handler }` and let `createRetryTaskOrDefault` do the single normalization.

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
```diff
@@ -530,22 +507,68 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
...
+      this._pendingTasks[id] = retryableTask;
+      return retryableTask;
+    }
+
+    if (options?.retry && isRetryHandler(options.retry)) {
+      // Normalize to AsyncRetryHandler — wraps sync handlers via Promise.resolve,
+      // and is a no-op for handlers that already return a Promise.
+      const handler = toAsyncRetryHandler(options.retry);
+      const retryHandlerTask = new RetryHandlerTask<TOutput>(
+        handler,
```
torosent: **Memory / Design (Info): Circular reference between `RetryHandlerTask` and `RuntimeOrchestrationContext`**

Here `this` (the `RuntimeOrchestrationContext`) is passed to the `RetryHandlerTask`, which is then stored in `this._pendingTasks[id]`. This creates a circular reference: `ctx._pendingTasks[id]._orchestrationContext === ctx`.

Modern V8 handles circular references fine with mark-and-sweep GC, so this isn't a leak, but it does mean the entire context (with all its state, pending tasks, received events, etc.) is pinned as long as any `RetryHandlerTask` reference exists — and vice versa. Both are short-lived (per-execution), so this is acceptable in practice.

Worth noting for awareness if the lifecycle of either object ever changes.

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
```diff
@@ -555,7 +578,7 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
       // Reschedule a sub-orchestration task
       const subOrch = originalAction.getCreatesuborchestration();
       if (!subOrch) {
-        throw new Error("Expected CreateSubOrchestrationAction on sub-orchestration retryable task");
+        throw new Error("Expected CreateSubOrchestrationAction on sub-orchestration retry task");
```
torosent: **Correctness (Medium): Sub-orchestration retry reuses the same instance ID**

When rescheduling a sub-orchestration retry, this code reuses `subOrch.getInstanceid()` from the original action. This relies on the backend supporting re-creation of a failed instance with the same ID.

If the Durable Task Scheduler backend treats instance IDs as unique and doesn't clean up failed instances before re-creation, retries could fail with a "duplicate instance" error. Please validate this behavior against the DTS backend, or consider generating a new deterministic instance ID for each retry attempt (e.g., appending the attempt count).

packages/durabletask-js/src/worker/orchestration-executor.ts
YunchuWang: done

packages/durabletask-js/src/task/retry-handler-task.ts
```diff
@@ -0,0 +1,89 @@
...
+    const totalRetryTimeMs = currentTime.getTime() - this.startTime.getTime();
+
+    const retryContext = createRetryContext(
+      this._orchestrationContext,
+      this.attemptCount,
+      failureDetails,
+      totalRetryTimeMs,
+      false, // isCancelled - not yet supported
+    );
+
```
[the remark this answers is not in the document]
YunchuWang: removed

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
YunchuWang: make sense, we can impose limits on these tasks timeout and retry counts to shorten the reference life though

packages/durabletask-js/src/task/options/task-options.ts
YunchuWang: removed this function.

packages/durabletask-js/src/worker/runtime-orchestration-context.ts
torosent: **Bug (Medium): Version field dropped on retry rescheduling**

In `rescheduleRetryTask()`, the `version` field from the original action is not extracted or passed when creating new actions for retry.

Compare with the original scheduling:
- `callActivity` (line 297): passes `options?.version` ✅
- `callSubOrchestrator` (line 328): passes `options?.version` ✅

But in `rescheduleRetryTask`:
- Activity (line 580): `ph.newScheduleTaskAction(newId, name, input, tags)` — missing version ❌
- Sub-orch (line 591): `ph.newCreateSubOrchestrationAction(newId, name, instanceId, input, tags)` — missing version ❌

For versioned activities/sub-orchestrations, retries would lose version constraint and could be picked up by the wrong worker version.

Suggested fix:

```typescript
// For activity:
const version = scheduleTask.getVersion()?.getValue();
newAction = ph.newScheduleTaskAction(newId, name, input, tags, version);

// For sub-orchestration:
const version = subOrch.getVersion()?.getValue();
newAction = ph.newCreateSubOrchestrationAction(newId, name, instanceId, input, tags, version);
```

packages/durabletask-js/src/worker/orchestration-executor.ts
```diff
@@ -564,4 +548,58 @@ export class OrchestrationExecutor {
...
+      }
+    } else if (task instanceof RetryHandlerTask) {
+      task.recordFailure(errorMessage, failureDetails);
+      const retryResult = await task.shouldRetry(ctx._currentUtcDatetime);
+
+      if (retryResult !== false) {
+        WorkerLogs.retryingTask(this._logger, ctx._instanceId, task.taskName, task.attemptCount);
+        task.incrementAttemptCount();
+
+        if (typeof retryResult === "number") {
```
torosent: **Robustness (Low): Handler returning `0` or negative delay crashes orchestration**

When a `RetryHandler` returns `0`, the flow goes through `typeof retryResult === "number"` → `ctx.createRetryTimer(task, retryResult)` → which throws `Error("delayMs must be a positive number, but received 0")`. This unhandled error propagates up and fails the entire orchestration.

The `RetryHandlerResult` JSDoc correctly says "a positive number" but there's no defensive validation in `tryHandleRetry`. A handler bug returning 0 would cryptically crash the orchestration.

Consider:
1. Treating `0` as "retry immediately" (use `rescheduleRetryTask` instead), or
2. Clamping to minimum 1ms, or
3. Adding a guard:
```typescript
if (typeof retryResult === "number") {
  if (retryResult <= 0) {
    // Treat non-positive delay as immediate retry
    ctx.rescheduleRetryTask(task);
  } else {
    ctx.createRetryTimer(task, retryResult);
  }
}
```

torosent (commented):
## Second Review — Post-Update Analysis

All 12 issues from the first review have been addressed — excellent responsiveness! The retry handler delay support (`RetryHandlerResult = boolean | number`) is well-designed and the code quality has improved significantly.

**4 new findings** in this follow-up:

| # | Severity | Finding |
|---|----------|----------|
| 1 | Medium | `isNonRetriable` flag not checked in `RetryableTask` (only checked in `RetryHandlerTask`) |
| 2 | Medium | Version field dropped when rescheduling retry tasks |
| 3 | Low | Handler returning `0` crashes orchestration instead of graceful handling |
| 4 | Low | Missing integration test for delay-based retry handler path |

packages/durabletask-js/src/worker/orchestration-executor.ts
YunchuWang: done

packages/durabletask-js/src/task/retryable-task.ts
```diff
@@ -165,14 +59,27 @@ export class RetryableTask<T> extends CompletableTask<T> {
    * The delay is capped at maxRetryInterval.
    */
   computeNextDelayInMilliseconds(currentTime: Date): number | undefined {
+    // Check if handleFailure predicate says we should NOT retry this failure type
+    if (this.lastFailure) {
```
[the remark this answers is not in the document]
YunchuWang: done

PR #154, opened 2026-03-10
fix: Destroy gRPC stream in getOrchestrationHistory to prevent resource leak

### LINKED ISSUE #153: [copilot-finds] Bug: gRPC stream not destroyed in getOrchestrationHistory causes resource leak
## Problem

`TaskHubGrpcClient.getOrchestrationHistory()` (`packages/durabletask-js/src/client/client.ts`, lines 955–975) calls `stream.removeAllListeners()` in both the `end` and `error` event handlers but does **not** call `stream.destroy()`. This leaves the underlying gRPC stream resources (TCP connections, buffers) unreleased after the stream completes.

In long-running applications that call `getOrchestrationHistory()` repeatedly (e.g., polling for orchestration status, monitoring dashboards), this can lead to:
- Connection pool exhaustion
- Memory leaks from accumulated stream objects
- Eventual resource starvation

## Root Cause

The cleanup pattern was incomplete. Only `removeAllListeners()` was called, which detaches event handlers but does **not** release the underlying Node.js stream resources. The `destroy()` method is required to fully clean up the stream.

## Evidence

The **worker code** (`task-hub-grpc-worker.ts`, lines 370–379, 443–444) already follows the correct pattern:

```typescript
stream.removeAllListeners();
stream.destroy();
```

The client code was missing the `stream.destroy()` call in both exit paths.

## Proposed Fix

Add `stream.destroy()` after `stream.removeAllListeners()` in both the `end` and `error` event handlers of `getOrchestrationHistory()`, matching the pattern used in the worker.

## Impact

**Severity:** Medium — affects applications that make frequent `getOrchestrationHistory()` calls.
**Scenarios affected:** Monitoring/polling workflows, orchestration history inspection tools, any code that repeatedly queries history.

YunchuWang:
## Summary

Fixes #153

**Bug:** \getOrchestrationHistory()\ in \client.ts\ calls \stream.removeAllListeners()\ in both the \nd\ and \rror\ event handlers but does **not** call \stream.destroy()\. This leaves the underlying gRPC stream resources (TCP connections, buffers) unreleased after the stream completes, causing potential connection pool exhaustion and memory leaks in long-running applications.

The worker code already follows the correct cleanup pattern (\
emoveAllListeners()\ + \destroy()\); this fix aligns the client code to match.

## Changes

- **\packages/durabletask-js/src/client/client.ts\** — Added \stream.destroy()\ after \stream.removeAllListeners()\ in both the \nd\ and \rror\ event handlers of \getOrchestrationHistory()\
- **\packages/durabletask-js/test/client-stream-cleanup.spec.ts\** — Added tests verifying stream cleanup on success and error paths

## Testing

All tests pass. Lint clean.

### DIFF
--- a/packages/durabletask-js/src/client/client.ts
+++ b/packages/durabletask-js/src/client/client.ts
@@ -954,11 +954,13 @@ export class TaskHubGrpcClient {
 
       stream.on("end", () => {
         stream.removeAllListeners();
+        stream.destroy();
         resolve(historyEvents);
       });
 
       stream.on("error", (err: grpc.ServiceError) => {
         stream.removeAllListeners();
+        stream.destroy();
         // Return empty array for NOT_FOUND to be consistent with DTS behavior
         // (DTS returns empty stream for non-existent instances) and user-friendly
         if (err.code === grpc.status.NOT_FOUND) {
--- a/packages/durabletask-js/test/client-stream-cleanup.spec.ts
+++ b/packages/durabletask-js/test/client-stream-cleanup.spec.ts
@@ -1,0 +1,185 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { EventEmitter } from "events";
+import * as grpc from "@grpc/grpc-js";
+import * as pb from "../src/proto/orchestrator_service_pb";
+import { TaskHubGrpcClient } from "../src/client/client";
+import { Timestamp } from "google-protobuf/google/protobuf/timestamp_pb";
+
+/**
+ * Creates a mock gRPC readable stream backed by an EventEmitter.
+ * Tracks whether destroy() was called.
+ */
+function createMockStream() {
+  const emitter = new EventEmitter();
+  const mockStream = Object.assign(emitter, {
+    destroyed: false,
+    destroy: jest.fn(() => {
+      mockStream.destroyed = true;
+    }),
+    cancel: jest.fn(),
+    getPeer: jest.fn(() => "localhost:4001"),
+  });
+  return mockStream;
+}
+
+/**
+ * Creates a mock gRPC stub with streamInstanceHistory returning the given mock stream.
+ */
+function createMockStub(mockStream: ReturnType<typeof createMockStream>) {
+  return {
+    streamInstanceHistory: jest.fn(() => mockStream),
+    close: jest.fn(),
+  };
+}
+
+/**
+ * Creates a HistoryChunk proto with a single ExecutionStarted event.
+ */
+function createMockHistoryChunk(): pb.HistoryChunk {
+  const chunk = new pb.HistoryChunk();
+  const event = new pb.HistoryEvent();
+  event.setEventid(1);
+  const ts = new Timestamp();
+  ts.fromDate(new Date());
+  event.setTimestamp(ts);
+  const executionStarted = new pb.ExecutionStartedEvent();
+  executionStarted.setName("TestOrchestration");
+  event.setExecutionstarted(executionStarted);
+  chunk.setEventsList([event]);
+  return chunk;
+}
+
+describe("TaskHubGrpcClient stream cleanup", () => {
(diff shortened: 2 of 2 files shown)


YunchuWang:
## Verification Finding: This fix is unnecessary — closing as not-a-bug

After thorough investigation, the resource leak described in this PR **does not exist**. The explicit \stream.destroy()\ calls are redundant because Node.js already handles this automatically. Detailed evidence below.

### Root Cause of the False Positive

The PR assumes that without an explicit \destroy()\ call, gRPC stream resources (TCP connections, buffers) will leak. This assumption is **incorrect** for the following reasons:

### Evidence 1: \utoDestroy\ defaults to \	rue\ (Node.js >= 14)

The Node.js official documentation for [\
ew stream.Readable([options])\](https://nodejs.org/api/stream.html#new-streamreadableoptions) states:

> \utoDestroy\ \<boolean\> Whether this stream should automatically call \.destroy()\ on itself after ending. **Default: \	rue\**.

This means **any Readable stream automatically calls \destroy()\ after the \'end'\ event fires**, releasing all internal resources. No explicit \destroy()\ call is needed.

### Evidence 2: gRPC-js does NOT disable \utoDestroy\

The \@grpc/grpc-js\ (v1.14.3) \ClientReadableStreamImpl\ constructor ([source: grpc-node/packages/grpc-js/src/call.ts#L110-L118](https://github.com/grpc/grpc-node/blob/main/packages/grpc-js/src/call.ts#L110-L118)):

\\\	ypescript
class ClientReadableStreamImpl<ResponseType> extends Readable {
  constructor(readonly deserialize) {
    super({ objectMode: true });  // Only sets objectMode, does NOT set autoDestroy: false
  }
}
\\\

Since \utoDestroy\ is not explicitly set to \alse\, it retains its default value of \	rue\.

### Evidence 3: gRPC-js always calls \push(null)\ before emitting \'error'\

In grpc-node's \client.ts\ [\makeServerStreamRequest\](https://github.com/grpc/grpc-node/blob/main/packages/grpc-js/src/client.ts#L603-L618), the \onReceiveStatus\ callback **always** calls \stream.push(null)\ before emitting \'error'\:

\\\	ypescript
onReceiveStatus(status: StatusObject) {
  receivedStatus = true;
  stream.push(null);  // ← EOF signal, triggers 'end', then autoDestroy kicks in
  if (status.code !== Status.OK) {
    stream.emit('error', callErrorFromStatus(status, callerStack));
  }
  stream.emit('status', status);
}
\\\

This means even on error paths, the stream receives the EOF signal, which triggers autoDestroy.

### Evidence 4: \
emoveAllListeners()\ does NOT prevent \utoDestroy\

Verified locally on Node.js v24.11.0:

\\\javascript
const {Readable} = require('stream');
const r = new Readable({objectMode: true, read(){}});
r.on('data', () => {});
r.push('hello');
r.push(null);
r.on('end', () => {
  r.removeAllListeners();  // ← This does NOT prevent autoDestroy
  console.log('after removeAllListeners');
});
setTimeout(() => console.log('destroyed=', r.destroyed), 100);
// Output: after removeAllListeners
// Output: destroyed= true  ← autoDestroy still works!
\\\

### Evidence 5: This project requires Node.js >= 22

\package.json\ specifies \ngines: { node: '>=22.0.0' }\. The \utoDestroy: true\ default has been in effect since Node.js 14 (added in [nodejs/node#30623](https://github.com/nodejs/node/pull/30623)), so it is guaranteed to be active.

### Regarding the worker code 'consistency' argument

The worker code (\	ask-hub-grpc-worker.ts\) also has explicit \destroy()\ calls that are equally redundant for the same reasons. Using one redundant pattern as justification to add another does not constitute a valid technical argument.

### Conclusion

- The \destroy()\ calls are **harmless** (it's an idempotent no-op when called after autoDestroy)
- But they are **unnecessary** — the claimed resource leak does not exist
- The PR description's claims about 'connection pool exhaustion' and 'memory leaks' are unfounded for this code path

Closing as not-a-bug.

PR #165, opened 2026-03-11
fix: Add missing non-Task validation in resume() isFailed path

### LINKED ISSUE #161: [copilot-finds] Bug: Missing non-Task validation in resume() isFailed path causes silent generator state corruption
## Problem

In `RuntimeOrchestrationContext.resume()` (`packages/durabletask-js/src/worker/runtime-orchestration-context.ts`, lines 148-166), when the previous task has failed and the generator catches the thrown exception, there is no validation that the generator yields a `Task` object.

If the orchestrator catches an exception and yields a non-Task value:

```typescript
async function* myOrchestrator(ctx) {
  try {
    yield ctx.callActivity(someActivity);
  } catch {
    yield "not a task"; // Bug: should fail with clear error
  }
}
```

The code silently falls through, leaving `_previousTask` pointing to the old failed task. On the next `resume()` call, the same exception is thrown to the generator in an unexpected location, causing confusing behavior.

## Root Cause

The `isComplete` path (line 184-186) correctly validates the generator output:

```typescript
if (!(value instanceof Task)) {
  throw new Error("The orchestrator generator yielded a non-Task object");
}
```

The `isFailed` path (lines 159-166) lacks this same validation:

```typescript
if (throwResult.value instanceof Task) {
  this._previousTask = throwResult.value;
  if (this._previousTask.isComplete) {
    await this.resume();
  }
  return;
}
// <-- Falls through silently when value is not a Task
```

## Proposed Fix

Add the same non-Task validation error after the `instanceof Task` check in the `isFailed` path, matching the existing validation in the `isComplete` path.

## Impact

**Severity: Medium.** When an orchestrator yields a non-Task value after catching an exception from a failed task:
- Instead of a clear "yielded a non-Task object" error, the orchestration enters an inconsistent state
- On the next event, the same exception is re-thrown to the generator in an unexpected location
- The resulting error message is confusing and does not indicate the actual problem (the non-Task yield)

This primarily affects orchestrators that catch task failures and have a programming error in their catch handler.

YunchuWang:
## Summary
Add non-Task validation in the isFailed path of 
esume(), making it consistent with the isComplete path.

Closes #161

## Problem
In 
esume(), the isComplete path correctly validates that the yielded value is a Task instance and throws a clear error. The isFailed path lacks this validation entirely - it silently returns when the yielded value is not a Task, leaving _previousTask pointing to the old failed task and causing confusing behavior on subsequent events.

## Fix
Add 	hrow new Error('The orchestrator generator yielded a non-Task object') after the instanceof Task check in the isFailed path.

### DIFF
--- a/packages/durabletask-js/src/worker/runtime-orchestration-context.ts
+++ b/packages/durabletask-js/src/worker/runtime-orchestration-context.ts
@@ -167,6 +167,8 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
           return;
         }
 
+        // The generator yielded a non-Task value after catching the exception.
+        // This is a user error — orchestrators must only yield Task objects.
         throw new Error("The orchestrator generator yielded a non-Task object");
       } else if (this._previousTask.isComplete) {
         while (true) {
--- a/packages/durabletask-js/test/orchestration_executor.spec.ts
+++ b/packages/durabletask-js/test/orchestration_executor.spec.ts
@@ -345,6 +345,90 @@ describe("Orchestration Executor", () => {
     expect(injectedTask.getResult()).toEqual("sub-result-zero");
   });
 
+  it("should fail the orchestration when the generator yields a non-Task value after catching an exception", async () => {
+    const dummyActivity = async (_: ActivityContext) => {
+      // do nothing
+    };
+    // This orchestrator catches the exception and yields a non-Task value (a string).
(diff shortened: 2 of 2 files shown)


YunchuWang:

YunchuWang:
Closing — the core fix was already merged via PR #163 (commit 5b58b2a). The remaining diff is only comments; duplicate.

PR #176, opened 2026-03-14
fix: Update instance status on suspend/resume in InMemoryOrchestrationBackend

### LINKED ISSUE #175: [copilot-finds] Bug: InMemoryOrchestrationBackend suspend/resume methods do not update instance status
## Problem

The `InMemoryOrchestrationBackend.suspend()` and `resume()` methods queue the appropriate `SuspendEvent`/`ResumeEvent` history events but never update `instance.status`. This causes `getOrchestrationState()` to return incorrect runtime status:

- After calling `suspend()`, status remains `RUNNING` instead of `SUSPENDED`
- After calling `resume()`, status remains `SUSPENDED` instead of `RUNNING`

**Files:** `packages/durabletask-js/src/testing/in-memory-backend.ts`, lines 168–203

## Root Cause

In the real DTS sidecar, the suspend/resume RPCs immediately transition the orchestration status. The executor does not produce a `COMPLETEORCHESTRATION` action for suspend/resume — it only sets an internal `_isSuspended` flag to buffer/unbuffer events. The in-memory backend relies on `processCompleteOrchestrationAction` to update status, but since no completion action is generated for suspend/resume, the status was never updated.

## Proposed Fix

1. Set `instance.status = ORCHESTRATION_STATUS_SUSPENDED` in `suspend()` to match real sidecar behavior
2. Transition `instance.status` from `SUSPENDED` back to `RUNNING` in `resume()`
3. Call `notifyWaiters()` from both methods so `waitForState()` predicates checking for `SUSPENDED` status are properly triggered

## Impact

- **Severity:** Medium — affects testing accuracy. Tests using the in-memory backend observe incorrect `RUNNING` status after suspend, which differs from real sidecar behavior.
- **Affected scenarios:** Any test checking `OrchestrationStatus.SUSPENDED` via `getOrchestrationState()` or `waitForState()` after calling `suspend()`.
- **Risk:** Low — change is scoped to the test infrastructure (`InMemoryOrchestrationBackend`), not the production gRPC path.

YunchuWang:
Fixes #175

## Problem
The InMemoryOrchestrationBackend suspend() and resume() methods queue the appropriate SuspendEvent/ResumeEvent history events but never update instance.status. This causes getOrchestrationState() to return incorrect runtime status.

## Changes
- Set instance.status = ORCHESTRATION_STATUS_SUSPENDED in suspend()
- Transition instance.status from SUSPENDED back to RUNNING in resume()
- Call notifyWaiters() from both methods so waitForState() predicates are properly triggered

### DIFF
--- a/packages/durabletask-js/src/testing/in-memory-backend.ts
+++ b/packages/durabletask-js/src/testing/in-memory-backend.ts
@@ -174,6 +174,10 @@ export class InMemoryOrchestrationBackend {
       throw new Error(`Orchestration instance '${instanceId}' not found`);
     }
 
+    if (this.isTerminalStatus(instance.status)) {
+      return; // Cannot suspend a completed/failed/terminated instance
+    }
+
     if (instance.status === pb.OrchestrationStatus.ORCHESTRATION_STATUS_SUSPENDED) {
       return;
     }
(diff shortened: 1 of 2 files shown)


YunchuWang:

YunchuWang:
Addressed all 3 Copilot review comments:

1. ✅ **Terminal status guard in suspend()** — Added `isTerminalStatus()` early return before setting SUSPENDED
2. ✅ **Terminal/non-suspended guard in resume()** — Added early returns for terminal status and non-SUSPENDED instances (resume is now a true no-op when not suspended)
3. ✅ **Missing waiter notification test for resume** — Added test that verifies `waitForState()` resolves on resume

Also added:
- Test: suspend on a completed instance is a no-op
- Test: resume on a non-suspended (RUNNING) instance is a no-op

kaibocai (approved):

PR #177, opened 2026-03-14
fix: Prevent TestOrchestrationWorker crash when completeOrchestration fails in error handler

### LINKED ISSUE #174: [copilot-finds] Bug: TestOrchestrationWorker processing loop crashes when completeOrchestration fails in error handler
## Problem

`InMemoryOrchestrationBackend.completeOrchestration()` throws an `Error` when the target instance does not exist (e.g., after `backend.reset()` or `purgeOrchestration()`), but the sibling method `completeActivity()` silently returns for the same scenario. This inconsistency causes `TestOrchestrationWorker.processOrchestration()` to crash when its catch block calls `completeOrchestration` on a purged instance.

**File:** `packages/durabletask-js/src/testing/in-memory-backend.ts`, line 257
**File:** `packages/durabletask-js/src/testing/test-worker.ts`, lines 155-158

## Root Cause

In `in-memory-backend.ts`:
- `completeOrchestration()` (line 257): `throw new Error(...)` when instance not found
- `completeActivity()` (line 311): silently returns when instance not found

In `test-worker.ts`, `processOrchestration()` has a catch block that calls `completeOrchestration` to report failures. If the instance was deleted between scheduling and completion (e.g., via `reset()`), the throw propagates out of the catch block, crashing `runProcessingLoop()`. The worker silently stops processing all subsequent work items.

## Proposed Fix

1. **Root cause:** Change `completeOrchestration` to return silently when the instance is not found, consistent with `completeActivity`.
2. **Defense-in-depth:** Wrap the catch block's `completeOrchestration` call in a nested try-catch so the worker survives even if `completeOrchestration` throws for other reasons.

## Impact

- **Severity:** Medium — affects testing infrastructure only (not production runtime)
- **Scenarios affected:** Any test that calls `backend.reset()` or `purgeOrchestration()` while orchestrations are in flight. The worker stops processing silently, causing subsequent test assertions to hang or time out with confusing errors.

YunchuWang:
Fixes #174

## Problem
InMemoryOrchestrationBackend.completeOrchestration() throws an Error when the target instance does not exist, but completeActivity() silently returns. This inconsistency causes TestOrchestrationWorker.processOrchestration() to crash when its catch block calls completeOrchestration on a purged instance.

## Changes
- Change completeOrchestration to return silently when instance not found, consistent with completeActivity
- Add defense-in-depth: wrap the catch block's completeOrchestration call in a nested try-catch so the worker survives

### DIFF
--- a/packages/durabletask-js/src/testing/in-memory-backend.ts
+++ b/packages/durabletask-js/src/testing/in-memory-backend.ts
@@ -256,7 +256,7 @@ export class InMemoryOrchestrationBackend {
   ): void {
     const instance = this.instances.get(instanceId);
     if (!instance) {
-      throw new Error(`Orchestration instance '${instanceId}' not found`);
+      return; // Instance may have been purged or the backend reset
     }
 
     if (instance.completionToken !== completionToken) {
--- a/packages/durabletask-js/src/testing/test-worker.ts
+++ b/packages/durabletask-js/src/testing/test-worker.ts
@@ -155,7 +155,12 @@ export class TestOrchestrationWorker {
         failureDetails,
       );
 
-      this.backend.completeOrchestration(instanceId, completionToken, [failAction]);
+      try {
+        this.backend.completeOrchestration(instanceId, completionToken, [failAction]);
+      } catch {
+        // Instance may have been purged or the backend reset during processing.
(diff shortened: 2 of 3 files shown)


YunchuWang:

PR #182, opened 2026-03-14
fix: Add stream cleanup and retry logic to gRPC error handler

### LINKED ISSUE #167: [copilot-finds] Bug: gRPC stream error handler does not recover, causing silent worker hang
## Problem

The `stream.on("error", ...)` handler in `TaskHubGrpcWorker.internalRunWorker()` (`packages/durabletask-js/src/worker/task-hub-grpc-worker.ts`, line 384) only logs the error but does **not** clean up the stream or attempt to reconnect:

```typescript
stream.on("error", (err: Error) => {
  if (this._stopWorker) {
    return;
  }
  WorkerLogs.streamErrorInfo(this._logger, err);
  // ← No cleanup, no retry
});
```

In contrast, the adjacent `stream.on("end", ...)` handler (line 370) correctly calls `removeAllListeners()`, `destroy()`, and `_createNewClientAndRetry()`.

## Root Cause

In Node.js with `@grpc/grpc-js`, gRPC stream errors — especially transport-level failures like `UNAVAILABLE` or abrupt network disconnections — may emit an `"error"` event **without** a subsequent `"end"` event. When this happens, the worker logs the error and then silently stops processing work items forever, because no recovery path is triggered.

The `stop()` method (line 424) already accounts for this asymmetry by listening for `"end"`, `"close"`, **or** `"error"` — confirming the developers know that `"error"` can fire alone.

## Proposed Fix

Add stream cleanup and retry logic to the error handler, mirroring the `"end"` handler pattern:
1. Call `stream.removeAllListeners()` to prevent double-recovery if both events fire
2. Add a no-op `stream.on("error", () => {})` guard to prevent unhandled error crashes from stale events
3. Call `stream.destroy()` to clean up the stream
4. Call `_createNewClientAndRetry()` to reconnect

## Impact

**Severity: High** — Affected workers silently stop processing orchestration/activity/entity work items after a transport-level gRPC error. This can happen in production when network connectivity is temporarily lost, load balancers reset connections, or the sidecar restarts. The worker appears healthy (no crash, no error logged at error level) but is effectively dead.

YunchuWang:
Fixes #167

## Problem
The stream.on('error', ...) handler in TaskHubGrpcWorker.internalRunWorker() only logs the error but does not clean up the stream or attempt to reconnect. In Node.js with @grpc/grpc-js, gRPC stream errors may emit an 'error' event without a subsequent 'end' event, causing the worker to silently stop processing work items forever.

## Changes
- Add stream cleanup and retry logic to the error handler, mirroring the 'end' handler pattern
- Call stream.removeAllListeners() to prevent double-recovery
- Add no-op error guard to prevent unhandled error crashes from stale events
- Call stream.destroy() and _createNewClientAndRetry() to reconnect

### DIFF
--- a/.git-status-check.txt
+++ b/.git-status-check.txt
@@ -1,2 +1,0 @@
-LATEST: test: rewrite e2e stream recovery as single Docker-based test
-?? .git-status-check.txt


YunchuWang:
## Addressing Copilot Review Comment: async Event Handlers

The Copilot reviewer correctly identified that `stream.on("error", async ...)` causes unhandled promise rejections. This has been fixed in commit 59a1ce7.

### What was wrong

The original code used `async` functions as EventEmitter listeners:

```typescript
stream.on("error", async (err: Error) => {
  // ...
  await this._createNewClientAndRetry();
});
```

### Why this is problematic (per Node.js official docs)

**1. Return values are discarded:**
> "Any values returned by the called listeners are **ignored and discarded**." — [Node.js Events docs](https://nodejs.org/api/events.html#events)

The Promise returned by an `async` listener is silently dropped. Nobody awaits it.

**2. Listeners are called synchronously:**
> "The EventEmitter calls all listeners **synchronously** in the order in which they were registered." — [Asynchronous vs. synchronous](https://nodejs.org/api/events.html#asynchronous-vs-synchronous)

`emit()` calls the listener, gets back a Promise, ignores it, and returns `boolean` synchronously.

**3. Node.js explicitly warns about this pattern:**
> "Using async functions with event handlers is **problematic**, because it can lead to an **unhandled rejection** in case of a thrown exception" — [Capture rejections](https://nodejs.org/api/events.html#capture-rejections-of-promises)

Even `captureRejections: true` does NOT make `emit()` await the Promise — it only routes rejections to an error handler as a safety net.

### What was fixed

Both `"end"` and `"error"` handlers were changed from `async` to synchronous, using `.catch()` for fire-and-forget error handling:

```typescript
stream.on("error", (err: Error) => {
  // ... synchronous cleanup ...
  this._createNewClientAndRetry().catch((retryErr) => {
    if (!this._stopWorker) {
      WorkerLogs.workerError(this._logger, retryErr instanceof Error ? retryErr : new Error(String(retryErr)));
    }
  });
});
```

Additionally, both handlers now include `stream.on("error", () => {})` after `removeAllListeners()` to prevent unhandled error events from crashing the process after cleanup (the `"end"` handler was previously missing this guard).

### Tests added
- 3 new unit tests covering retry rejection handling and the end-handler error guard
- 1 comprehensive e2e test that starts a worker with no emulator, verifies retries via structured log capture, then starts the Docker emulator and verifies the worker reconnects and completes an orchestration

PR #256, opened 2026-06-11
fix: preserve entity executor stack traces

### LINKED ISSUE #248: [copilot-finds] Bug: Entity executor silently loses stack traces due to protobuf setValue() return value misuse
## Problem

In `packages/durabletask-js/src/worker/entity-executor.ts` line 399, the entity executor's error handling uses:

```typescript
failureDetails.setStacktrace(new StringValue().setValue(error.stack));
```

In google-protobuf, `setValue()` returns `void` (not the StringValue instance for fluent chaining). This means `setStacktrace()` receives `undefined`, silently discarding the stack trace for **all** entity operation failures.

## Root Cause

The code assumes `setValue()` returns `this` for method chaining, but google-protobuf setters return `void`. This is a one-character-class-of-bug: the code compiles and runs without error, but the stack trace is never actually set on the failure details.

## Proposed Fix

Construct the StringValue separately and pass it to `setStacktrace()`, matching the correct pattern used in `pb-helper.util.ts` (lines 200-202):

```typescript
const stackValue = new StringValue();
stackValue.setValue(error.stack);
failureDetails.setStacktrace(stackValue);
```

## Impact

**Severity: Medium** — Every entity operation failure loses its stack trace, making debugging entity errors significantly harder. Users see error type and message but cannot trace back to the exact line that threw. This affects all entity-based workflows using the V1 batch execution path.

YunchuWang:
## Summary
- constructs StringValue separately before setting entity failure stack traces
- preserves stack traces for entity operation failures
- adds regression coverage for entity executor failure details

## Tests
- Added/updated entity executor tests in packages/durabletask-js/test/entity-executor.spec.ts

Fixes #248

### DIFF
--- a/packages/durabletask-js/src/worker/entity-executor.ts
+++ b/packages/durabletask-js/src/worker/entity-executor.ts
@@ -396,7 +396,9 @@ export class TaskEntityShim {
         failureDetails.setErrortype(error.name);
         failureDetails.setErrormessage(error.message);
         if (error.stack) {
-          failureDetails.setStacktrace(new StringValue().setValue(error.stack));
+          const stackValue = new StringValue();
+          stackValue.setValue(error.stack);
+          failureDetails.setStacktrace(stackValue);
         }
       } else {
         failureDetails.setErrortype("Error");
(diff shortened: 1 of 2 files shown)


YunchuWang:
Closing after local verification showed the premise is invalid: google-protobuf StringValue#setValue() returns the StringValue instance in the package version used by packages/durabletask-js, not void/undefined, so stack traces are not discarded by this chaining pattern.

PR #274, opened 2026-06-18
fix: reject non-finite retry policy values

### LINKED ISSUE #230: [copilot-finds] Bug: RetryPolicy constructor silently accepts NaN and Infinity, enabling infinite retries
## Problem

`RetryPolicy` validates constructor parameters using comparison operators (`<=`, `<`) that silently accept `NaN` and `Infinity` because JavaScript NaN comparisons always return `false`.

**File:** `packages/durabletask-js/src/task/retry/retry-policy.ts`, lines 67-94

For example:
```typescript
// NaN <= 0 evaluates to false, so NaN passes this validation!
if (maxNumberOfAttempts <= 0) {
  throw new Error("maxNumberOfAttempts must be greater than zero");
}
```

## Root Cause

JavaScript NaN comparison semantics: `NaN <= 0`, `NaN < 1.0`, and `NaN >= x` all return `false`. This means every comparison-based guard in the constructor is bypassed when NaN is passed.

This can happen when users compute retry parameters from external sources:
```typescript
const maxAttempts = parseInt(process.env.MAX_RETRIES!); // NaN if env var is undefined
```

## Impact

- **`maxNumberOfAttempts: NaN`** — Most severe. The check `attemptCount >= maxNumberOfAttempts` in `RetryableTask.computeNextDelayInMilliseconds()` always returns `false` when `maxNumberOfAttempts` is `NaN`, causing **infinite retries** that never stop.
- **`firstRetryIntervalInMilliseconds: NaN`** — Produces `NaN` timer delays, creating timers with `Invalid Date`.
- **`backoffCoefficient: NaN`** — Produces `NaN` delay values on all retry calculations.
- **`Infinity` values** — Similar issues: `Infinity` maxNumberOfAttempts allows unlimited retries, `Infinity` intervals create timers that cannot fire.

## Proposed Fix

Add `Number.isFinite()` guards before all comparison checks in the `RetryPolicy` constructor. `Number.isFinite()` returns `false` for `NaN`, `Infinity`, and `-Infinity`, catching all invalid numeric values.

YunchuWang:
## Summary
- Rejects NaN and Infinity numeric retry policy parameters to avoid invalid retry behavior.
- Opens the existing fix branch for issue #230

Fixes #230

### DIFF
--- a/packages/durabletask-js/src/task/retry/retry-policy.ts
+++ b/packages/durabletask-js/src/task/retry/retry-policy.ts
@@ -63,6 +63,8 @@ export class RetryPolicy {
       retryTimeoutInMilliseconds,
       handleFailure,
     } = options;
+    const resolvedMaxRetryIntervalInMilliseconds = maxRetryIntervalInMilliseconds ?? 3600000;
+    const resolvedRetryTimeoutInMilliseconds = retryTimeoutInMilliseconds ?? -1;
 
     // Validation aligned with .NET SDK
     // Use !Number.isFinite() guards to reject NaN and Infinity, which bypass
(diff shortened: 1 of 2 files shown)


packages/durabletask-js/src/task/retry/retry-policy.ts
```diff
@@ -63,43 +63,46 @@ export class RetryPolicy {
       retryTimeoutInMilliseconds,
       handleFailure,
     } = options;
+    const resolvedMaxRetryIntervalInMilliseconds = maxRetryIntervalInMilliseconds ?? 3600000;
```
YunchuWang, line 66: why  default to 3600000

packages/durabletask-js/src/task/retry/retry-policy.ts
```diff
@@ -63,43 +63,46 @@ export class RetryPolicy {
...
     }
 
     if (
-      maxRetryIntervalInMilliseconds !== undefined &&
-      maxRetryIntervalInMilliseconds !== -1 &&
-      maxRetryIntervalInMilliseconds < firstRetryIntervalInMilliseconds
+      resolvedMaxRetryIntervalInMilliseconds !== -1 &&
+      (!Number.isFinite(resolvedMaxRetryIntervalInMilliseconds) ||
+        resolvedMaxRetryIntervalInMilliseconds < firstRetryIntervalInMilliseconds)
     ) {
```
YunchuWang, line 88: whats this for

kaibocai (approved):

PR #282, opened 2026-06-29
Add `durable-functions@4.0.0` — Azure Functions Durable provider on the gRPC core (+ core host helpers, E2E CI, and release pipeline)

YunchuWang:
## TL;DR for reviewers

This PR began as "add two Functions gRPC helpers to the core worker" (the original title), but it now delivers the **full Azure Functions Durable provider on the gRPC core**. The headline change is a **new npm package `durable-functions@4.0.0`** (`packages/azure-functions-durable/`, ~30 source files — does **not** exist on `main`) that re-implements the classic Durable Functions programming model on top of `@microsoft/durabletask-js`. It is the gRPC-core successor to the classic **v3** (extension‑HTTP) `durable-functions` package and targets the `@azure/functions` **v4** programming model. The core worker helpers, the E2E suite enablement, and the release pipeline are the supporting cast.

**Scope: 58 files, +4,648 / −523, across four areas:**

| # | Area | What | Key paths |
|---|------|------|-----------|
| 1 | **New provider package** (the bulk) | `durable-functions@4.0.0` on the gRPC core — v3‑compatible programming model | `packages/azure-functions-durable/**` |
| 2 | Core host helpers (original scope) | Single‑work‑item byte processors on `TaskHubGrpcWorker` | `packages/durabletask-js/src/worker/task-hub-grpc-worker.ts`, `src/index.ts` |
| 3 | E2E suite + CI | Wire the (pre‑existing) Functions host E2E suite to run green on PRs | `.github/workflows/functions-e2e-tests.yaml`, `test/e2e-functions/**` |
| 4 | Build & release | Pack + ESRP‑publish the new package | `eng/templates/build.yml`, `eng/ci/release.yml` |

---

## 1. New package: `durable-functions@4.0.0` (`packages/azure-functions-durable/`)

A v3‑API‑compatible Azure Functions Durable provider built on the gRPC core, so existing v3 orchestrator/activity/entity/client code type‑checks and runs against the consolidated gRPC path. Added whole in this PR (not on `main`).

- **Registration surface** (`app.ts`, `app-client.ts`, `trigger.ts`, `input.ts`, `worker.ts`): `app.orchestration` / `app.activity` / `app.entity`, the `app.client.*` durable‑client starter helpers (HTTP / timer / blob / queue / service‑bus / event‑hub / event‑grid / cosmos), and trigger + input bindings.
- **Client** (`client.ts`, 616 lines): `DurableFunctionsClient` / `DurableOrchestrationClient` — start, terminate, suspend, resume, raiseEvent, getStatus, purge, rewind, restart. Includes `_mapControlPlaneError`, which translates opaque gRPC control‑plane errors back to the v3 error surface (see **Error fidelity** below).
- **Execution contexts** (`orchestration-context.ts`, `entity-context.ts`): `DurableOrchestrationContext` / `DurableEntityContext` plus `wrapOrchestrator` / `wrapEntity` adapters bridging the classic generator model to the core.
- **Status / query / types**: `orchestration-status.ts`, `orchestration-filter.ts`, `metadata.ts`, `retry-options.ts`, `entity-id.ts`, `entity-state-response.ts`, `purge-history-result.ts`, `http-management-payload.ts`, `get-client.ts`, `durable-grpc.ts`, `user-agent.ts`.
- **Top‑level v3 compat aliases**: re‑exports `TaskFailedError` from core for `instanceof` guards; type aliases (`DurableClient`, `EntityContext<T>`, `OrchestrationHandler`, …) so v3 code compiles unchanged. `README.md` + `CHANGELOG.md` document migration notes and intentional omissions (`DurableError` / `AggregatedError`).
- **~15 unit spec files** (~2,000 lines) covering client, client‑query, app, app‑client, orchestration‑context, entity‑context, query‑types, status, retry‑options, trigger, worker, get‑client, compat‑exports, durable‑grpc, entity‑id — **91/91 passing**.

## 2. Core host helpers on `TaskHubGrpcWorker` (original PR scope)

Two methods that let a host drive **one** durable work item per invocation without the long‑lived gRPC worker loop — mirroring the Python consolidation in durabletask-python #155:

- `processOrchestratorRequest(request: Uint8Array): Promise<Uint8Array>`
- `processEntityBatchRequest(request: Uint8Array): Promise<Uint8Array>`

Each deserializes a `TaskHubSidecarService` protobuf, executes one work item via the worker's existing internal path (`_executeOrchestratorInternal` / `_executeEntityInternal`) using an in‑process `CapturingSidecarStub` (null‑stub pattern, same as Python's `AzureFunctionsNullStub`), and returns the serialized response. Transport encoding (e.g. base64) stays in the host, out of core. Covered by `packages/durabletask-js/test/functions-grpc-support.spec.ts`.

## 3. E2E suite + CI enablement

The Functions host E2E suite already exists on `main`; this PR makes it actually run — and pass — against the new provider:

- **Workflow** (`.github/workflows/functions-e2e-tests.yaml`): runs on `pull_request` for the Functions surface (`test/e2e-functions/**`, `packages/azure-functions-durable/**`, the workflow file) plus `workflow_dispatch`; adds `--skipApiVersionCheck` to Azurite; fixes build ordering; and **fixes a latent invalid‑YAML step name that was blocking the workflow entirely** (including manual dispatch).
- **Test app** (`test/e2e-functions/test-app/host.json`): targets the Preview extension bundle `[4.29.0, 5.0.0)` (gRPC path) and sets `maxGrpcMessageSizeInBytes: 2147483647` to lift the host's default 4 MiB gRPC receive cap (which otherwise fails large‑output / get‑all‑instances). See **Cross‑language note**.
- **Specs / harness**: `harness.ts` expected‑error strings aligned to the v4 gRPC‑core fidelity; `suspend-resume.spec.ts` / `terminate.spec.ts` updated, with 3 terminal‑instance control‑plane cases skipped pending #315.
- Docs / runner: `README.md`, `test/e2e-functions/README.md`, `scripts/test-e2e-functions.sh`.
- **Result: 32 passed / 0 failed / 13 skipped** (the 13 = 3 new #315 skips + pre‑existing environment/feature skips).

## 4. Build & release pipeline

- `eng/templates/build.yml`: pack `durable-functions` and stage the `azure-functions-durable` artifact.
- `eng/ci/release.yml`: new `durable_functions` ESRP release job to publish the package.

---

## Error fidelity (`_mapControlPlaneError`)

The consolidated gRPC path collapses control‑plane failures into two opaque codes; the compat layer translates them back to the v3 surface:

- `NOT_FOUND` (5) → `No instance with ID '<id>' found.` (matches v3).
- `UNKNOWN` (2) on a **non‑terminal** instance → looks up runtime status and throws `Cannot <op> orchestration instance in the <State> state.` (matches host wording).
- `UNKNOWN` (2) on a **terminal** instance → detail is not recoverable from the opaque error; the v4 status policy (200‑swallow vs 409/410, ideally a server‑side fix) is deferred to **#315**, so those 3 specs are skipped.

Activity/entity failure text now surfaces the v4‑core shape (`TaskFailedError: Activity task #N failed: …`, `EntityOperationFailedException: …`) rather than v3's `Activity function '<name>' failed:`. The core error does not carry the function name, and reconstructing the exact v3 string would require a replay‑risky change in the shared core, so the E2E expectations were aligned to the v4‑native shape. (Name‑in‑error is a possible future core enhancement benefiting all SDKs.)

## Cross‑language note

The 4 MiB gRPC cap is an extension‑defined default (.NET `DurableTaskOptions.MaxGrpcMessageSizeInBytes = 4194304`), sent to and honored consistently by every SDK worker. Raising it via `maxGrpcMessageSizeInBytes` in `host.json` is the same knob .NET‑isolated and Python use — no SDK‑side divergence.

## Follow‑ups

- **#315** — terminal/invalid‑state control‑plane ops return an opaque `2 UNKNOWN`; decide the v4 status policy (server‑side gRPC status‑detail fix preferred). 3 E2E specs skipped pending this.
- *(Optional)* decouple the JS client's send vs receive gRPC caps for exact .NET‑isolated parity.

## Testing

- Compat unit: **91/91** (`npm run test:unit -w durable-functions`).
- Core: `functions-grpc-support.spec.ts` green.
- Functions host E2E: **32 passed / 0 failed / 13 skipped** (Preview bundle + Azurite).
- `npm run build -w durable-functions` clean (core + compat).


packages/azure-functions-durable/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+## 4.0.0-alpha.0
+
+- Added the initial gRPC-consolidated Azure Functions Durable provider package.
```
YunchuWang: too detailed, just one bullet point

packages/azure-functions-durable/README.md
```diff
@@ -0,0 +1,130 @@
+# durable-functions
```
YunchuWang, line 1: rewrite this readme, i dont need the implementation plan. just tell whats this package support for, why it is needed.

packages/azure-functions-durable/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "url": "https://github.com/microsoft/durabletask-js/issues"
+  },
+  "homepage": "https://github.com/microsoft/durabletask-js/tree/main/packages/azure-functions-durable#readme",
+  "engines": {
+    "node": ">=22.0.0"
+  },
+  "dependencies": {
+    "@azure/functions": "^4.16.1",
+    "@grpc/grpc-js": "^1.14.4",
+    "@microsoft/durabletask-js": "0.3.0"
```
YunchuWang: TODO: remember to bump this to 0.4.0, and we need to release @microsoft/durabletask-js first

also confirm whether @azure/functions 4.16.1 has the extension changes required,

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
```diff
@@ -313,6 +313,58 @@ export class TaskHubGrpcWorker {
...
+   * Host integrations own any transport-specific encoding (for example base64).
+   */
+  async processOrchestratorRequest(request: Uint8Array): Promise<Uint8Array> {
+    const req = pb.OrchestratorRequest.deserializeBinary(request);
+    const stub = new CapturingSidecarStub();
+    await this._executeOrchestratorInternal(req, "", stub as unknown as stubs.TaskHubSidecarServiceClient);
+
+    if (!stub.orchestratorResponse) {
+      throw new Error("Orchestrator execution did not produce a response.");
+    }
```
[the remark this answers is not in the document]
YunchuWang: Checked closely — the abandon branch is **unreachable in the Azure Functions (Durable extension) path**, so keeping the generic error is fine.

`_executeOrchestratorInternal` only abandons inside the version-mismatch branch, which requires **all** of:
1. `this._versioning` set with `matchStrategy !== None` — otherwise the version check short-circuits to compatible ([L590](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/durabletask-js/src/worker/task-hub-grpc-worker.ts#L590))
2. a non-empty worker version ([L599](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/durabletask-js/src/worker/task-hub-grpc-worker.ts#L599))
3. an incompatible incoming orchestration version, **and**
4. `failureStrategy !== Fail` (i.e. Reject) so `shouldFail === false` ([L637-L639](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/durabletask-js/src/worker/task-hub-grpc-worker.ts#L637-L639)) → the abandon branch ([L740-L757](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/durabletask-js/src/worker/task-hub-grpc-worker.ts#L740-L757))

The Functions path never configures versioning: the shared worker is built as `new DurableFunctionsWorker()` with no options ([app.ts#L22](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/azure-functions-durable/src/app.ts#L22)), and `DurableFunctionsWorker` just calls `super(options)`. So `_versioning` is `undefined`, `_checkVersionCompatibility` returns `{ compatible: true }` at [L590](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/durabletask-js/src/worker/task-hub-grpc-worker.ts#L590), and the `if (!versionCheckResult.compatible)` block (incl. abandon) is skipped entirely.

Abandon is also structurally a worker-loop concept (reject a work item so *another* worker picks it up); the single-work-item host model has no second worker, and the extension owns dispatch/retry. The capturing stub already no-ops abandon by design, matching the Python provider's null stub ([L1245-L1254](https://github.com/microsoft/durabletask-js/blob/00d705e0d5699bb11ed284fdcbe519635ef18809/packages/durabletask-js/src/worker/task-hub-grpc-worker.ts#L1245-L1254)).

So `processOrchestratorRequest` throwing "did not produce a response" is only reachable via a genuine executor bug (neither a completion nor an exception), which is the right thing to surface. No change needed unless we later expose versioning to the Functions worker.

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
```diff
@@ -313,6 +313,58 @@ export class TaskHubGrpcWorker {
...
+   * single orchestration work item per invocation instead of running the
+   * long-lived gRPC worker loop. It reuses the same execution path as the worker
+   * loop, capturing the response in-process rather than completing it over gRPC.
+   * Host integrations own any transport-specific encoding (for example base64).
+   */
+  async processOrchestratorRequest(request: Uint8Array): Promise<Uint8Array> {
+    const req = pb.OrchestratorRequest.deserializeBinary(request);
+    const stub = new CapturingSidecarStub();
+    await this._executeOrchestratorInternal(req, "", stub as unknown as stubs.TaskHubSidecarServiceClient);
+
```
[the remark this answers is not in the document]
YunchuWang: By design for this single-work-item host path. `processOrchestratorRequest` is a byte-level helper for the Azure Functions flow, where the host owns work-item dispatch and completion-token correlation over its own gRPC sidecar; the helper just returns the executed `OrchestratorResponse` bytes. This mirrors the Python provider, which likewise does not stamp a completion token at this layer. If we ever drive `completeOrchestratorTask` directly from JS, we can thread the token through then. Resolving.

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
```diff
@@ -313,6 +313,58 @@ export class TaskHubGrpcWorker {
...
+   * single entity batch work item per invocation instead of running the
+   * long-lived gRPC worker loop. It reuses the same execution path as the worker
+   * loop, capturing the result in-process rather than completing it over gRPC.
+   * Host integrations own any transport-specific encoding (for example base64).
+   */
+  async processEntityBatchRequest(request: Uint8Array): Promise<Uint8Array> {
+    const req = pb.EntityBatchRequest.deserializeBinary(request);
+    const stub = new CapturingSidecarStub();
+    await this._executeEntityInternal(req, "", stub as unknown as stubs.TaskHubSidecarServiceClient);
+
```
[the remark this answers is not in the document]
YunchuWang: Same rationale as the orchestrator helper: by design for the single-work-item host path. `processEntityBatchRequest` returns the executed `EntityBatchResult` bytes and leaves completion-token correlation to the host's gRPC sidecar, matching the Python provider. We can thread a token through if we ever call `completeEntityTask` directly from JS. Resolving.

packages/azure-functions-durable/CHANGELOG.md
YunchuWang: Done in 00d705e — collapsed the `4.0.0-alpha.0` section to a single bullet.

packages/azure-functions-durable/README.md
YunchuWang: Done in 00d705e — rewrote the README around what the package supports and why it is needed, and dropped the implementation-plan / phase-status / open-questions sections.

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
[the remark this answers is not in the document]
YunchuWang: Confirmed — abandon is unreachable in the Functions path (no versioning configured on the shared worker), and abandon has no meaning in the single-work-item host model where the extension owns dispatch/retry. Keeping the generic "did not produce a response" error is correct, since it only surfaces a genuine executor bug. Resolving.

packages/azure-functions-durable/src/orchestration-context.ts
```diff
@@ -0,0 +1,263 @@
...
+ * @remarks
+ * Core-native orchestrators declare `(ctx, input)` and are returned unchanged. Classic v3
+ * orchestrators declare a single `context` parameter and use `context.df.*`; those are wrapped so
+ * the core engine drives them while `context.df` forwards to the core {@link OrchestrationContext}.
+ *
+ * Detection is by arity: only single-parameter functions are treated as classic. Declare your
+ * core-native orchestrator as `(ctx, input)` so it passes through.
+ */
+export function wrapOrchestrator(handler: TOrchestrator | ClassicOrchestrator): TOrchestrator {
+  if (typeof handler === "function" && handler.length === 1) {
```
andystaples: **Bug (high): classic-vs-native detection by arity silently mis-runs native single-arg orchestrators.**

`handler.length === 1` treats *any* one-parameter function as classic. But a core-native orchestrator written as `async function*(ctx) { yield ctx.callActivity(...) }` — single param, ignoring `input`, a very common style — also has arity 1. It gets wrapped and invoked as `classic(classicCtx)`, where `classicCtx` is `{ df, log, ... }`, so `ctx.callActivity` is `undefined`.

Worse: because `isGenerator()` (line 257) checks for `[Symbol.iterator]`, it returns `false` for the *async* generator the handler returns, so the wrapper `return`s the un-iterated generator and the engine completes the orchestration immediately with garbage output — **the body never runs and nothing throws**. (Note `functions-grpc-support.spec.ts` itself writes a native orchestrator as single-arg `async (_ctx) => ...`.)

Suggest detecting by generator kind instead of arity, which matches the engine's actual async-iterator gate: `handler.constructor.name === "GeneratorFunction"` ⇒ classic (v3 sync generators); `AsyncGeneratorFunction` / async ⇒ native pass-through. Please also add a round-trip test that drives a native single-arg orchestrator through `wrapOrchestrator` → the real executor.

packages/azure-functions-durable/src/orchestration-context.ts
```diff
@@ -0,0 +1,263 @@
...
+  ): Task<T> {
+    return this._ctx.callSubOrchestrator<unknown, T>(name, input, {
+      retry: retryOptions.toRetryPolicy(),
+      instanceId,
+      version,
+    });
+  }
+
+  /** Creates a durable timer that fires at the specified time. */
+  createTimer(fireAt: Date | number): Task<unknown> {
```
andystaples: **Break: `createTimer` returns a non-cancelable core `Task`, so the canonical durable-timeout pattern no longer works.**

v3 `createTimer` returned a `TimerTask` with `.cancel()` / `.isCanceled`. The ubiquitous pattern

```js
const timeout = context.df.createTimer(deadline);
const work = context.df.callActivity("DoWork");
const winner = yield context.df.Task.any([timeout, work]);
if (!timeout.isCompleted) timeout.cancel();
```

breaks two ways: core `Task` has no `cancel()` (→ `TypeError`, or a leaked timer that keeps the instance alive until it fires), and the result shape changed (`isCompleted/isFaulted/result` → `isComplete/isFailed/getResult()`), so `timeout.isCompleted` is always `undefined`. This is one of the most common DF patterns — consider surfacing a cancelable timer wrapper, or calling it out prominently in the migration notes.

packages/azure-functions-durable/src/orchestration-context.ts
```diff
@@ -0,0 +1,263 @@
...
+    entityId: EntityInstanceId,
+    operationName: string,
+    operationInput?: unknown,
+  ): void {
+    this._ctx.entities.signalEntity(entityId, operationName, operationInput);
+  }
+}
+
+/** The object passed to a classic (v3) orchestrator function; its `df` is the durable context. */
+export interface ClassicOrchestrationContext {
```
andystaples, line 180: **Break: the classic context no longer extends `InvocationContext`.**

v3's `OrchestrationContext extends InvocationContext`, so orchestrators could read `context.invocationId`, `context.functionName`, `context.extraInputs`, etc. This interface exposes only `df` + log methods, so any orchestrator touching those members won't compile/run. Same applies to `ClassicEntityContext` in `entity-context.ts` (just `{ df }`). Worth documenting as a breaking change.

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,502 @@
...
+    this.managementUrls = config.managementUrls ?? {};
+    this.baseUrl = config.baseUrl ?? "";
+    this.requiredQueryStringParameters = requiredQueryStringParameters;
+    this.rpcBaseUrl = rpcBaseUrl;
+    this.httpBaseUrl = config.httpBaseUrl ?? "";
+    this.maxGrpcMessageSizeInBytes = maxGrpcMessageSizeInBytes;
+    this.grpcHttpClientTimeout = config.grpcHttpClientTimeout;
+  }
+
+  createCheckStatusResponse(request: HttpRequest, instanceId: string): HttpResponse {
```
andystaples: **Subtle break:** v3's signature was `createCheckStatusResponse(request: HttpRequest | undefined, instanceId)` and fell back to the binding's `baseUrl` when `request` was undefined. This overload requires a defined `request` and will throw at `request.url` (in `getInstanceStatusUrl`) if called with `undefined`. Consider accepting `HttpRequest | undefined` to preserve the v3 contract.

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,502 @@
...
+
+  /**
+   * Gets the status of an orchestration instance in the classic Durable Functions (v3) shape.
+   *
+   * @deprecated Use {@link getOrchestrationState} instead.
+   * @param instanceId - The ID of the orchestration instance to query.
+   * @param options - When `showInput` is `false`, input/output payloads are not fetched.
+   * @returns The instance status, or `undefined` if the instance does not exist.
+   */
+  async getStatus(
```
andystaples: **Subtle break:** the return type is `DurableOrchestrationStatus | undefined` where v3 returned a non-optional `DurableOrchestrationStatus`, so existing callers doing `(await getStatus(id)).runtimeStatus` now get a TS error / possible runtime throw. Also `GetStatusOptions.showHistory` / `showHistoryOutput` are silently ignored and `history` is never populated (only `showInput` is honored). Worth documenting the reduced option set — and note `startNew` similarly drops v3's `version` option.

packages/azure-functions-durable/src/orchestration-status.ts
```diff
@@ -0,0 +1,127 @@
...
+    input: parseJson(state.serializedInput),
+    output: parseJson(state.serializedOutput),
+    runtimeStatus: toOrchestrationRuntimeStatus(state.runtimeStatus),
+    customStatus: parseJson(state.serializedCustomStatus),
+  });
+}
+
+/** @hidden */
+function parseJson(value: string | undefined): unknown {
+  return value === undefined ? undefined : JSON.parse(value);
```
andystaples: **Nit:** `parseJson` guards `undefined` but not `""`. `JSON.parse("")` throws, so a present-but-empty serialized payload would blow up `getStatus` / `toDurableOrchestrationStatus`. Normal flow yields `undefined` (absent `StringValue` wrapper) so this is low-risk, but cheap to harden: `if (!value) return undefined;`.

packages/azure-functions-durable/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  DurableOrchestrationStatusInit,
+  OrchestrationRuntimeStatus,
+  fromOrchestrationRuntimeStatus,
+  toDurableOrchestrationStatus,
+  toOrchestrationRuntimeStatus,
+} from "./orchestration-status";
+export { EntityId } from "./entity-id";
+export { EntityStateResponse } from "./entity-state-response";
+export { PurgeHistoryResult } from "./purge-history-result";
+export { OrchestrationFilter } from "./orchestration-filter";
```
andystaples, line 48: **Break: several v3 top-level exports are missing**, so `import { ... } from "durable-functions"` breaks for:

- `DummyOrchestrationContext` / `DummyEntityContext` — v3 testing utilities; users' existing unit tests won't compile.
- `TaskFailedError` (and v3's `DurableError` / `AggregatedError`) — needed for `instanceof` guards on caught orchestration errors; core throws its own `TaskFailedError`, which isn't re-exported here.
- `ManagedIdentityTokenSource` / `TokenSource`.

Relatedly, the `app` surface drops the v3 durable-client trigger helpers (`df.app.client.http` / `timer` / `storageQueue` / ...) and `setExceptionPropertiesProvider`. The replacement model (generic `app.http` + `df.input.durableClient()` + `df.getClient(context)`) is reasonable, but it changes client-function handler signatures and should be documented as a breaking migration.

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
```diff
@@ -313,6 +313,58 @@ export class TaskHubGrpcWorker {
...
+   * long-lived gRPC worker loop. It reuses the same execution path as the worker
+   * loop, capturing the response in-process rather than completing it over gRPC.
+   * Host integrations own any transport-specific encoding (for example base64).
+   */
+  async processOrchestratorRequest(request: Uint8Array): Promise<Uint8Array> {
+    const req = pb.OrchestratorRequest.deserializeBinary(request);
+    const stub = new CapturingSidecarStub();
+    await this._executeOrchestratorInternal(req, "", stub as unknown as stubs.TaskHubSidecarServiceClient);
+
+    if (!stub.orchestratorResponse) {
```
andystaples, line 335: **Minor (core):** if versioning is enabled and a version mismatch resolves to *abandon*, `_executeOrchestratorInternal` calls the stub's no-op `abandonTaskOrchestratorWorkItem` and returns without setting a response — so this throws a generic `"Orchestrator execution did not produce a response."`. Unreachable via the Functions path today (the worker is constructed without versioning), but since this is a public, reusable entry point, consider distinguishing "abandoned" from a genuine failure, or documenting that abandon isn't supported on the single-work-item path.

packages/durabletask-js/src/proto/orchestrator_service_pb.d.ts
```diff
@@ -1756,8 +1756,8 @@ export class OrchestratorRequest extends jspb.Message {
     static extensions: {[key: number]: jspb.ExtensionFieldInfo<jspb.Message>};
     static extensionsBinary: {[key: number]: jspb.ExtensionFieldBinaryInfo<jspb.Message>};
     static serializeBinaryToWriter(message: OrchestratorRequest, writer: jspb.BinaryWriter): void;
-    static deserializeBinary(bytes: Uint8Array): OrchestratorRequest;
     static deserializeBinaryFromReader(message: OrchestratorRequest, reader: jspb.BinaryReader): OrchestratorRequest;
+    static deserializeBinary(bytes: Uint8Array): OrchestratorRequest;
```
andystaples, line 1793: **Nit:** this looks like generated-proto churn — `deserializeBinary` / `deserializeBinaryFromReader` are merely reordered with no semantic change. Worth dropping from the diff to keep it focused (and confirm the proto codegen version isn't drifting).

packages/azure-functions-durable/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "grpc"
+  ],
+  "author": "Microsoft",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/microsoft/durabletask-js/issues"
+  },
+  "homepage": "https://github.com/microsoft/durabletask-js/tree/main/packages/azure-functions-durable#readme",
+  "engines": {
+    "node": ">=22.0.0"
```
andystaples, line 48: **Nit:** `engines.node >= 22` is a notable consumer-facing floor (drops the Node 18/20 that v3 supported) — worth calling out in the README/changelog. It's also inconsistent with `@types/node@^18` in `devDependencies`.

andystaples (commented):
## Review summary

Two things ship here: (1) the advertised **core gRPC byte-processors** (`processOrchestratorRequest` / `processEntityBatchRequest` + `CapturingSidecarStub`) in `task-hub-grpc-worker.ts`, and (2) an entire new **`durable-functions@4.0.0-alpha.0`** package (`packages/azure-functions-durable`) — the intended next-major replacement for `Azure/azure-functions-durable-js` v3.

The core helpers are clean, and reusing the existing execution path with an in-process capturing stub is a nice low-risk approach. The v3 compat adapters (`context.df.*` for orchestrations/entities and the client method surface) are genuinely broad and well documented.

My main concern is the **"drop-in / run unchanged" framing** in the README: this is a strong compatibility *shim* for the common method surface, but several heavily-used v3 patterns and the entire durable-client authoring model are not covered — and there's one silent-failure bug in the orchestrator adapter. Details are inline; grouped here:

**Bug (please fix)**
- `wrapOrchestrator` picks classic-vs-native by parameter **arity**, which silently mis-runs a native single-arg orchestrator (`async function*(ctx)`) — the body never executes and no error is thrown.

**Behavioral breaks vs v3 (worth closing or documenting)**
- `createTimer` returns a non-cancelable `Task`; the `Task.any` + `timeoutTask.cancel()` timeout pattern breaks, and the `Task` result shape changed (`isCompleted/isFaulted/result` → `isComplete/isFailed/getResult()`).
- Orchestration/entity contexts no longer extend `InvocationContext`.
- Missing top-level exports: `DummyOrchestrationContext`/`DummyEntityContext` (breaks users' tests), error types (`DurableError`/`AggregatedError`/`TaskFailedError`) for `instanceof`, `ManagedIdentityTokenSource`/`TokenSource`; the `app.client.*` durable-client trigger helpers and `setExceptionPropertiesProvider` are also gone.
- Client deltas: `getStatus` returns `| undefined` and ignores `showHistory`; `createCheckStatusResponse` now requires a defined `request`; `startNew` drops the `version` option.

**Nits**
- `parseJson` doesn't guard `""` (`JSON.parse("")` throws).
- Core: a version-mismatch *abandon* makes `processOrchestratorRequest` throw a generic error.
- Spurious `deserializeBinary` reorder in the generated proto `.d.ts`.
- `engines.node >= 22` vs `@types/node@^18`; the Node-22 floor is a notable consumer bump.

Overall a promising foundation — I'd fix the arity bug and soften/annotate the compatibility claims before this reads as a transparent v3 replacement. Leaving this as comments (non-blocking).

packages/durabletask-js/src/proto/orchestrator_service_pb.d.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/orchestration-status.ts
YunchuWang: addressed in f935bce -- empty-string already guarded; now also tolerant of non-JSON serializedOutput

packages/azure-functions-durable/src/client.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/index.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/client.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/package.json
YunchuWang: addressed in f935bce

packages/azure-functions-durable/package.json
YunchuWang: addressed in f935bce

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
YunchuWang: addressed in f935bce

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: addressed in 8f62a35 — refined: classic vs core-native orchestrators are now detected by generator/async kind instead of pure arity (async function*/async => core-native pass-through; sync function* => classic, wrapped; plain sync fn => arity fallback), and the wrapper drives both sync and async generators via yield*. Added end-to-end regression tests through the real core executor: single-arg native async-generator body actually runs and yields a task; classic single-arg sync generator still works; classic non-generator return value still works.

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: addressed in 8f62a35 — createTimer's return type is now TimerTask (core already returns a cancelable TimerTask via #293; only the type was hidden as Task<unknown>). Added a test covering .cancel()/.isCanceled type visibility plus the Task.any(timer, work) then timer.cancel() timeout-race pattern.

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: addressed in 8f62a35 (doc-only) — CHANGELOG + README migration sections now record the breaking change: classic orchestration context exposes only df + replay-safe log helpers (no invocationId/functionName/extraInputs), and the classic entity context is just { df }. Added the rationale (reading InvocationContext members inside an orchestrator is replay-nondeterministic and was never recommended). No members restored.

packages/azure-functions-durable/src/client.ts
YunchuWang: addressed in 8f62a35 — confirmed v3-faithful: createCheckStatusResponse(request: HttpRequest | undefined, instanceId) already accepts an undefined request and falls back to the durable-client binding's baseUrl to build the management payload; undefined-request test is in place. One nuance to flag: the compat layer reconstructs the status URL from baseUrl rather than rewriting the binding's managementUrls templates (behavior-equivalent for the status link on the consolidated path).

packages/azure-functions-durable/src/client.ts
YunchuWang: addressed in 8f62a35 — getStatus now matches v3: returns a non-optional DurableOrchestrationStatus and throws when the instance is missing (verified against v3 DurableClient.getStatus, which throws on the extension's 404). showHistory populates history from core getOrchestrationHistory; startNew forwards the v3 version option to the core scheduler (scheduleNewOrchestration supports version). Honest boundaries (documented in JSDoc + CHANGELOG/README): history entries are core HistoryEvent shape, not the .NET extension's history shape (gRPC path bypasses the extension HTTP formatter); showHistoryOutput is accepted but can't strip payloads (core events always carry them); showInput maps to core fetchPayloads, so showInput:false also gates output/customStatus (v3 omitted only input).

YunchuWang: further refined in 900064f — getStatus/startNew now implement all three v3 points (not just documented):

(a) not-found: verified v3 DurableClient.getStatus (src/durableClient/DurableClient.ts) — its case 404 throws and the return type is non-optional Promise<DurableOrchestrationStatus>. Replicated: returns non-optional status, throws a 'DurableClient error: No orchestration instance with ID ... was found.' Error when the instance is missing.

(b) showHistory/showHistoryOutput: showHistory populates history via core getOrchestrationHistory; showHistoryOutput toggles per-entry payloads (strips input/result when falsy, keeps them when true). v3 types history as Array<unknown> (types/orchestration.d.ts), so entries are core HistoryEvent[]; the .NET extension's PascalCase HTTP history shape isn't reproduced because the gRPC path bypasses that formatter. showInput now gates only the top-level input (output/customStatus always returned, matching v3).

(c) version: startNew forwards options.version to core scheduleNewOrchestration (StartOrchestrationOptions.version).

Tests updated: not-found throw, showInput gates input only, showHistoryOutput strip vs keep, startNew version. build + lint + compat unit (81/81) green.

YunchuWang: not-found message aligned with v3 in 0dcc47c — mirrors v3 DurableClient.getStatus's exact sentence: 'This usually means we could not find any data associated with the instanceId provided: <id>.' Only v3's HTTP-404-specific first sentence is replaced (the gRPC path has no HTTP 404). Behavior is otherwise unchanged from 900064f: non-optional return + throw on not-found, showHistory/showHistoryOutput, showInput gates only input, startNew forwards version.

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
YunchuWang: Final state for this thread: the guardrail — a distinct, actionable error on the single-work-item `processOrchestratorRequest` abandon path, recorded via `CapturingSidecarStub.abandoned` — landed in f935bce and is unchanged. The explanatory comment was later made self-contained in cc420b8 (removed an inaccurate cross-SDK parity aside). Keeping the JS-specific diagnostic here is intentional.

packages/durabletask-js/src/proto/orchestrator_service_pb.d.ts
YunchuWang: Confirmed reverted — the generated proto file (`orchestrator_service_pb.d.ts`) is no longer part of this PR's diff (base..HEAD shows no `src/proto/*` change), so there is no reordering churn and no codegen-version drift introduced here. Resolving.

packages/azure-functions-durable/src/index.ts
YunchuWang: Decision: keeping the removed-and-documented approach rather than adding compat shims.
- `TaskFailedError` is re-exported from the core SDK (matches v3's top-level `TaskFailedError`).
- `DurableError` / `AggregatedError` are intentionally not provided — the core engine surfaces `TaskFailedError` plus JS-native `AggregateError`.
- func testing api in v3 is different from v4 which does not accept history events and context to run orch directly. added a separate github issue for v4 testing api work
- `ManagedIdentityTokenSource` / `TokenSource` are not part of the consolidated gRPC surface.
- The app-surface change (generic `app.http` + `df.input.durableClient()` + `df.getClient(context)`, and the dropped `df.app.client.*` / `setExceptionPropertiesProvider`) is documented as a breaking migration in CHANGELOG + README.
This keeps the compat surface aligned with the core (and the Python SDK) and avoids speculative shims. Resolving.

packages/azure-functions-durable/package.json
YunchuWang: Decision: the Node >= 22 floor is intended for the v4 consolidation — the `@microsoft/durabletask-js` core targets modern Node. The `@types/node@^18` inconsistency you flagged is fixed: devDependencies now use `@types/node@^22.0.0` (with `typescript@^5`). The floor is called out in both CHANGELOG ("Node.js >= 22 (drops the Node 18/20 that durable-functions v3 supported)") and the README migration section. Resolving.

packages/azure-functions-durable/src/client.ts
YunchuWang: Confirmed complete and v3-faithful: `createCheckStatusResponse(request: HttpRequest | undefined, instanceId)` accepts an undefined request and falls back to the durable-client binding's `baseUrl`. The base URL is wired from the host-provided binding through `DurableFunctionsClient` config parsing (`optionalString(record, "baseUrl")` -> `this.baseUrl` -> `getInstanceStatusUrl`). `createHttpManagementPayload` keeps both overloads — `(instanceId)` (v3 classic, baseUrl-based) and `(request, instanceId)` (request-origin) — both are public v3-compat API and are also used internally by `createCheckStatusResponse`. As in v3, the management-URL base comes from the binding; when present it is used verbatim. Resolving.

packages/azure-functions-durable/package.json
YunchuWang: Release-readiness findings (version bump to 0.4.0 is done in code; the two open items below are release-time actions):

**1) `@azure/functions` 4.16.1 floor — confirmed adequate.**
- v4.16.0 carried the substantive change: bumped `@azure/function-extensions-base` 0.2.0 → 0.3.0 and dropped Node 18 from CI (which lines up with this package's Node >= 22 floor).
- v4.16.1 is a security-servicing release over 4.16.0 (MSRC updates); v4.16.2 is functionally identical to 4.16.1 (only reclaimed the npm `latest` tag from an erroneously published 3.6.0).
- This package uses only stable generic-binding APIs (`trigger.generic` / `input.generic` carrying `durableRequiresGrpc`), available since v4.0, and does not depend on `@azure/function-extensions-base` directly. So `^4.16.1` is a sound "recent + security-patched" floor (it resolves to 4.16.2 today). The actual durable-gRPC enablement (`durableRequiresGrpc` → gRPC channel) is host-extension-side, not gated by the npm library version.

**2) Core release ordering — a real blocker to verify before publish.**
- The core package (`@microsoft/durabletask-js`) on this branch still declares `version: 0.3.0`, but it adds new public APIs this package relies on (`processOrchestratorRequest`, `addNamedOrchestrator`, `CapturingSidecarStub`, …). The published `0.3.0` on npm does NOT contain those.
- This package's dependency is pinned to `"@microsoft/durabletask-js": "0.3.0"`. Locally the monorepo workspace links the in-tree core (which has the new APIs), so E2E passes; but once published to npm, `0.3.0` would resolve to the old published core and fail at runtime (`processOrchestratorRequest is not a function`).
- Release checklist: (a) bump the core version (e.g. 0.4.0) and publish `@microsoft/durabletask-js` first; (b) update this package's dep from `0.3.0` to the new core version (`0.4.0` / `^0.4.0`); (c) then publish `durable-functions@0.4.0`.

Leaving this thread open until the core is released and the dependency is bumped.

packages/azure-functions-durable/CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
+# Changelog
```
YunchuWang, line 1: dont change changelog.md in this pr

packages/azure-functions-durable/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "durable-functions",
+  "version": "0.4.0",
```
andystaples: **Ship as `4.0.0`, not `0.4.0` (P0).** This package reuses the `durable-functions` npm name, whose current `latest` is **3.4.0**.

`0.4.0` sorts *below* 3.x, so consumers on a `^3` range never resolve it, and moving the `latest` dist-tag to `0.4.0` is a semver downgrade — it will not supplant v3 in place. A proper successor major is **`4.0.0`**.

(Core `@microsoft/durabletask-js` staying at `0.4.0` is correct — separate, pre-1.0 package.)

packages/azure-functions-durable/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "files": [
+    "dist",
+    "LICENSE",
+    "README.md",
+    "CHANGELOG.md"
+  ],
+  "scripts": {
+    "clean": "node -e \"require('fs').rmSync('dist', {recursive:true, force:true})\"",
+    "prebuild": "node -p \"'// Auto-generated by prebuild\\nexport const SDK_VERSION = ' + JSON.stringify(require('./package.json').version) + ';\\nexport const SDK_PACKAGE_NAME = ' + JSON.stringify(require('./package.json').name) + ';'\" > src/version.ts",
```
andystaples: **Dead prebuild (P3).** This generates `src/version.ts` (gitignored), but nothing imports it — `user-agent.ts` reads `../package.json` directly. Either consume `version.ts` or drop this `prebuild` step (and its call in `build`).

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,616 @@
...
+      try {
+        runtimeStatus = (await this.getStatus(instanceId)).runtimeStatus;
+      } catch {
+        runtimeStatus = undefined;
+      }
+      if (runtimeStatus !== undefined && !isTerminalRuntimeStatus(runtimeStatus)) {
+        throw new Error(`Cannot ${operationVerb} orchestration instance in the ${runtimeStatus} state.`);
+      }
+    }
+    throw error;
```
andystaples, line 532: **Terminal-state control-plane ops throw where v3 no-ops (P1).**

For an *already-terminal* instance, `isTerminalRuntimeStatus(runtimeStatus)` is true, so the L462 block is skipped and execution falls through to `throw error` here — rethrowing the opaque gRPC `UNKNOWN`. (Same outcome if the `getStatus` on L458 itself fails: `runtimeStatus` stays `undefined` and you fall through.)

v3 treats `terminate`/`suspend`/`resume` on a completed/failed instance as a **success no-op** (HTTP 410). This is the #315 gap — 3 E2E specs are skipped — and it is not in the CHANGELOG breaking-changes.

Recommend swallowing `UNKNOWN`-on-terminal to restore the v3 no-op, or documenting it prominently and unskipping the specs once #315 lands. (Python #155 also rethrows raw here — cross-SDK — so aligning both would be ideal.)

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,616 @@
...
+  }
+
+  /**
+   * Reads the state of a durable entity in the classic Durable Functions (v3) shape.
+   *
+   * @deprecated Use {@link getEntity} instead.
+   * @param entityId - The target entity instance ID.
+   * @param includeState - Whether to include the entity state in the response (default `true`).
+   */
+  async readEntityState<T = unknown>(entityId: EntityInstanceId, includeState = true): Promise<EntityStateResponse<T>> {
```
andystaples: **`readEntityState` signature diverges from v3 (P2).** v3 is `readEntityState(entityId, options?: TaskHubOptions)`; here the second param is `includeState: boolean`.

A v3 TS caller passing the options object no longer compiles (object vs boolean). Single-arg callers are unaffected, but the `taskHubName`/`connectionName` routing options are silently gone. Worth a CHANGELOG note.

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,616 @@
...
+   *   when falsy, those fields are stripped from each event; when `true`, they are kept.
+   */
+  async getStatus(instanceId: string, options?: GetStatusOptions): Promise<DurableOrchestrationStatus> {
+    const state = await this.getOrchestrationState(instanceId, true);
+    if (state === undefined) {
+      // Mirror v3's DurableClient.getStatus not-found behavior: it throws (its case-404 branch) with
+      // this message shape. The gRPC path has no HTTP 404, so only the extension-specific first
+      // sentence is replaced; the instanceId-bearing sentence is kept verbatim.
+      throw new Error(
+        `DurableClient error: No orchestration instance found. ` +
```
andystaples: **Minor (P3):** this reasonably rewrites the v3 not-found message but drops the literal `HTTP 404` substring (the instanceId sentence is preserved). Fine for humans, but anyone string-matching on `HTTP 404` will break — worth a CHANGELOG line.

packages/azure-functions-durable/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+// Licensed under the MIT License.
+
+import type { ClassicEntityContext, ClassicEntity } from "./entity-context";
+import type { DurableFunctionsClient } from "./client";
+
+export * as app from "./app";
+export * as trigger from "./trigger";
+export * as input from "./input";
+
+export {
```
andystaples, line 11: **Entity locking / critical sections dropped and undocumented (P1).** v3 exports `DurableLock`, `LockState`, and `LockingRulesViolationError`, and its orchestration context exposes `df.lock(...)` / `isLocked()`. None exist here, and the CHANGELOG removed-exports list does not mention them.

The core SDK already provides `context.entities.lockEntities()` / `LockHandle`, so this is bridgeable. Either implement `lock`/`isLocked` + re-export the types, or explicitly document the removal with migration guidance for customers using critical sections.

packages/azure-functions-durable/CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
...
+  `isCompleted` / `isFaulted` / `result`). `context.df.createTimer(...)` returns a cancelable
+  `TimerTask`, so the timeout-race pattern (`Task.any` then `timer.cancel()`) keeps working.
+- `client.getStatus()` matches the v3 signature: it returns a non-optional `DurableOrchestrationStatus`
+  and throws when the instance does not exist. `showInput` suppresses only the top-level input (output
+  and custom status are always returned, as in v3); `showHistory` populates `history`, and
+  `showHistoryOutput` toggles whether those entries keep their input/result payloads. One
+  consolidated-path note: `history` entries are core `HistoryEvent`s (v3 types `history` as
+  `Array<unknown>`), not the classic .NET extension's history serialization.
+- `client.startNew()` supports the v3 `version` option (forwarded to the core scheduler).
+- Removed top-level exports: `DummyOrchestrationContext`, `DummyEntityContext`, `DurableError`,
```
andystaples: **Removed-exports list is inaccurate (P2).** Verified against v3 `src/index.ts`: `DurableError`, `AggregatedError`, and `TokenSource` were **never** v3 top-level exports, so listing them as removed is misleading (only `ManagedIdentityTokenSource` and the two `Dummy*` contexts genuinely were).

Conversely, the genuinely-removed entity-lock exports (`DurableLock`, `LockState`, `LockingRulesViolationError`) and the now-throwing `callHttp` are **missing** from this list. Please reconcile it with the actual v3 surface.

packages/azure-functions-durable/src/orchestration-context.ts
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  /**
+   * Schedules a durable HTTP call (classic v3 API).
+   *
+   * @remarks
+   * Not supported: the durabletask engine has no durable-HTTP (`callHttp`) equivalent, so this
+   * throws. This mirrors the Python provider, which raises for the same reason.
+   */
+  callHttp(_options: unknown): never {
```
andystaples, line 156: **`callHttp` now throws (P3).** Matches the Python provider and is a reasonable scoping call, but durable HTTP (`callHttp`) is a v3 capability being removed — it should appear in the CHANGELOG breaking-changes so callers are not surprised at runtime.

andystaples (commented):
## Review summary

Thanks for this — it is a well-structured provider that closely mirrors the Python PR (Azure/durabletask-python#155) and gets a lot right: `RetryOptions`→`RetryPolicy` mapping, the `OrchestrationFilter` shape, `getStatus` throw-on-not-found, `startNew` `version` parity, and tolerant status parsing (`parseJson`). I verified locally — compat unit tests **91/91**, core `functions-grpc-support.spec.ts` **3/3** — and diffed the full public surface against the legacy `durable-functions` v3 package and the published core tarball.

Feedback is inline, grouped by priority:

- **P0 — blocks a clean in-place upgrade:** the package reuses the `durable-functions` npm name (v3 `latest` = 3.4.0) but ships as **`0.4.0`**; it needs to be **`4.0.0`** to actually supplant v3. *(Nice — the core `0.4.0` bump, matching dep, and release-order note already landed, so the earlier stale-core concern is resolved.)*
- **P1 — behavior a v3 customer hits in-place:** terminal-state `terminate`/`suspend`/`resume` now throw instead of v3 no-op; entity locking / critical sections (`df.lock`, `isLocked`, and the `DurableLock`/`LockState`/`LockingRulesViolationError` exports) are dropped and undocumented.
- **P2 — docs/signature accuracy:** the CHANGELOG removed-exports list is inaccurate/incomplete; `readEntityState` second parameter changed shape.
- **P3 — minor:** dead `version.ts` prebuild, `getStatus` error-string wording, and `callHttp` now throwing.

Nothing here is architectural — the design is sound. Details in the line comments.

packages/azure-functions-durable/src/client.ts
YunchuWang: restored

packages/azure-functions-durable/src/index.ts
YunchuWang: need change in core, tracking as separte issue for followup if it is doable, documenting as breaking change to use core apis

packages/azure-functions-durable/CHANGELOG.md
YunchuWang: removed, tracking in separate issue

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: Documented in the CHANGELOG breaking-changes section in abe250f — `context.df.callHttp(...) now throws`, with the host-vs-worker rationale (v3 ran it as a built-in host HTTP activity; the consolidated gRPC backend has no host-managed durable-HTTP primitive). Also filed #318 to track restoring it as a worker-side durable activity (fetch-based, phased: basic → 202-polling → managed-identity). Resolving since the documentation ask is addressed.

test/e2e-functions/test-app/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "test-app",
     "version": "1.0.0",
     "description": "",
-    "comment_durable-functions": "This app is wired to the PUBLISHED durable-functions package so the suite is runnable without an in-repo build. To test against an in-repo build of the Azure Functions durable package instead, change the durable-functions dependency below to a file: reference (e.g. file:../../../packages/azure-functions-durable) once that package exists on your branch, then re-run npm install.",
+    "comment_durable-functions": "This app is wired to the IN-REPO durable-functions (compat) package and core @microsoft/durabletask-js via file: references, so the E2E suite exercises the on-branch build. Build the in-repo packages first (npm run build -w durable-functions from the repo root), then run npm install here. The @microsoft/durabletask-js override is REQUIRED: the compat package pins core to a published version that predates the Functions worker helpers (processOrchestratorRequest/processEntityBatchRequest), so without the override npm would resolve stale published core and the worker path would fail at runtime.",
```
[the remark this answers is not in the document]
YunchuWang: Fixed in e3bd426 — reworded the `comment_durable-functions` rationale in the E2E test-app package.json. The override is required because the compat package depends on the on-branch, not-yet-published `@microsoft/durabletask-js@0.4.0` that *introduces* the worker helpers (processOrchestratorRequest/processEntityBatchRequest), not because core "predates" them.

packages/azure-functions-durable/CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
...
+  `AggregatedError`, `ManagedIdentityTokenSource`, `TokenSource`. `TaskFailedError` is re-exported
+  from the core SDK and aggregate failures surface as JS-native `AggregateError`. For orchestration
+  unit tests, use the core `TestOrchestrationWorker` / `TestOrchestrationClient` in place of the v3
+  dummy contexts.
+
+### Release order
+
+- Publish `@microsoft/durabletask-js` (0.3.0) before this package.
```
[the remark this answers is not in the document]
YunchuWang: Already addressed — the release-order note instructing to publish `@microsoft/durabletask-js` 0.3.0 first was removed from the CHANGELOG in an earlier revision. No stale release-order coupling remains.

README.md
```diff
@@ -5,7 +5,23 @@
 
 This repo contains a JavaScript/TypeScript SDK for use with the [Azure Durable Task Scheduler](https://github.com/Azure/Durable-Task-Scheduler). With this SDK, you can define, schedule, and manage durable orchestrations using ordinary TypeScript/JavaScript code.
 
-> Note that this SDK is **not** currently compatible with [Azure Durable Functions](https://learn.microsoft.com/azure/azure-functions/durable/durable-functions-overview). If you are looking for a JavaScript SDK for Azure Durable Functions, please see [this repo](https://github.com/Azure/azure-functions-durable-js).
+> Note that this SDK does **not** provide the [Azure Durable Functions](https://learn.microsoft.com/azure/azure-functions/durable/durable-functions-overview) programming model, decorators, or worker-indexing metadata. If you are looking for a JavaScript SDK for Azure Durable Functions, please see [this repo](https://github.com/Azure/azure-functions-durable-js). This package exposes low-level TaskHubSidecarService gRPC/protobuf helpers that host integrations can reuse; those helpers follow this package's Node.js 22+ requirement.
```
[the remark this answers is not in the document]
YunchuWang: Fixed in e3bd426 — the root README blockquote now points readers to the in-repo `durable-functions` provider package under `packages/azure-functions-durable` for the Azure Durable Functions programming model, and clarifies that the core `@microsoft/durabletask-js` package only exposes low-level gRPC/protobuf helpers. The classic v3 (extension-HTTP) predecessor link is retained.

packages/azure-functions-durable/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "durable-functions",
+  "version": "4.0.0",
+  "description": "Azure Functions Durable provider for the Durable Task JavaScript SDK",
```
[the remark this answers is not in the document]
YunchuWang: Already correct — the `durable-functions` package version is 4.0.0, and the PR title/description consistently refer to 4.0.0. The `@microsoft/durabletask-js: 0.4.0` entry is the core dependency version, which is intentionally independent of the compat package's 4.0.0.

packages/azure-functions-durable/src/get-client.ts
```diff
@@ -0,0 +1,52 @@
...
+ * constructor accepts either shape, so it is passed straight through.
+ *
+ * @param context - The invocation context for the current function call.
+ * @returns A rich Durable Functions client bound to the host's local gRPC sidecar.
+ * @throws If no `durableClient`/`orchestrationClient` input binding is registered, or its value
+ *   is not a valid client configuration.
+ */
+export function getClient(context: InvocationContext): DurableFunctionsClient {
+  const clientInput = context.options.extraInputs.find(isDurableClientInput);
+  if (!clientInput) {
```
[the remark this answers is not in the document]
YunchuWang: Fixed in e3bd426 — `getClient` now optional-chains `context.options.extraInputs?.find(...)`, so a missing `extraInputs` collection falls through to the friendly "no durable client input binding" error instead of throwing a raw TypeError.

packages/azure-functions-durable/src/app-client.ts
```diff
@@ -0,0 +1,265 @@
...
+    options.extraInputs.push(input.durableClient());
+  }
+}
+
+/** @hidden Wraps a client handler so the DurableFunctionsClient is injected as the 2nd argument. */
+function convertToFunctionHandler(clientHandler: DurableClientHandler): FunctionHandler {
+  return (triggerInput: unknown, context: InvocationContext): FunctionResult<any> => {
+    const client = getClient(context);
+    return clientHandler(triggerInput, client, context);
+  };
```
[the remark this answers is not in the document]
YunchuWang: Fixed in e3bd426 — `getClient` (called by `convertToFunctionHandler`) now reuses a `DurableFunctionsClient` and its underlying gRPC channel via a module-level cache keyed by the binding config, instead of constructing a new client/channel on every invocation. Invalid input still throws before anything is cached.

packages/azure-functions-durable/src/orchestration-context.ts
```diff
@@ -0,0 +1,296 @@
...
+    // orchestrators. A classic v3 orchestrator is never a plain async function — it must be a
+    // (sync) generator to `yield` durable tasks — so `async` unambiguously means core-native.
+    return false;
+  }
+  if (kind === "GeneratorFunction") {
+    return true; // classic v3: a sync generator the engine can't drive on its own.
+  }
+  // Plain SYNC (non-generator) function: kind is ambiguous, so fall back to arity. A lone
+  // `context` parameter is the classic v3 shape; `(ctx, input)` is core-native.
+  return handler.length <= 1;
```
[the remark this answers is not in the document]
YunchuWang: Tracked as #321 rather than expanded here (PR scope is core helpers + CI + release pipeline).

Confirmed this is a genuine classification tie, not a quick fix: a plain sync single-arg handler is the identical shape for a classic `(context) => context.df.getInput()` and a core-native `(ctx) => ctx.getInput()`, so no static kind/arity check can separate them. The compat also has an intentional test treating that shape as classic (orchestration-context.spec.ts "exposes a replay-safe logger as context.log/error on the classic context"), and v3 confirms a classic orchestrator is always a single-arg sync generator (TaskOrchestrationExecutor drives `fn(context)` via `.next()`), so flipping the arity default is a behavior reversal, not a one-liner. Severity is low — only a trivial, purely synchronous single-arg core-native orchestrator (no await/yield) is affected. Options and a recommended fix (explicit tagging at registration) are captured in #321.

packages/azure-functions-durable/src/entity-context.ts
```diff
@@ -0,0 +1,150 @@
...
+   * @remarks
+   * When the context is typed `EntityContext<TState>` (v3 style), a bare `getState()` returns
+   * `TState | undefined`. The per-call type parameter can still override this (e.g. `getState<X>()`).
+   *
+   * @param initializer - Optional zero-argument callable providing the initial state when none exists.
+   */
+  getState<T = TState>(initializer?: () => T): T | undefined {
+    const defaultValue = typeof initializer === "function" ? initializer() : undefined;
+    return this._operation.state.getState<T>(defaultValue);
+  }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in d78421e — `DurableEntityContext.getState` now short-circuits on `this._operation.state.hasState`: it returns the stored state via `getState<T>()` when state is present and only falls back to the initializer when there is none. This matches Durable Functions v3 (`Azure/azure-functions-durable-js` `Entity.getState`, which returns the existing state and invokes the initializer only in the no-state branch), so an expensive or side-effecting initializer no longer runs when state already exists. Added a regression test asserting the initializer is not called when state exists. Build green; unit 94/94.

packages/azure-functions-durable/src/get-client.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const bindingData: unknown = context.extraInputs.get(clientInput);
+  const cacheKey = typeof bindingData === "string" ? bindingData : JSON.stringify(bindingData);
+  let client = clientCache.get(cacheKey);
+  if (!client) {
+    client = new DurableFunctionsClient(asClientInput(bindingData));
+    clientCache.set(cacheKey, client);
+  }
+  return client;
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4561dc3. `getClient` now wraps the cached client's `stop()` so the cache entry is evicted on dispose — a later `getClient()` with the same binding builds a fresh client instead of handing back a closed gRPC channel. Added a unit test (`getClient` → `stop()` → `getClient` returns a new instance). `asClientInput` still throws before construction, so invalid input is never cached.

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,653 @@
...
+          // (`case 410: return;`). The consolidated gRPC path collapses that to an opaque UNKNOWN, so
+          // detect the terminal state and return without throwing.
+          return;
+        }
+        // Non-terminal wrong-state (e.g. suspend of an already-suspended instance): surface the
+        // host-side friendly wording instead of the opaque UNKNOWN.
+        throw new Error(`Cannot ${operationVerb} orchestration instance in the ${runtimeStatus} state.`);
+      }
+    }
+    throw error;
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4561dc3. `_mapControlPlaneError` now attaches the original gRPC error as `{ cause: error }` on every mapped throw ("No instance with ID ... found." and "Cannot <op> ... <State> state."), so the underlying status code/details/stack remain available for logging/debugging while callers still get the v3-compatible message.

packages/azure-functions-durable/CHANGELOG.md
YunchuWang: Done — `packages/azure-functions-durable/CHANGELOG.md` is reduced to a minimal `## TBD` placeholder in 0ef7ebb; the detailed 4.0.0 entries will be authored at release time, not in this PR. The root `CHANGELOG.md` was **not** modified by this PR (verified byte-identical to `origin/main`). Resolving per your direction.

packages/azure-functions-durable/src/client.ts
```diff
@@ -0,0 +1,677 @@
...
+    if (
+      operationVerb !== undefined &&
+      (code === grpcStatus.UNKNOWN || code === grpcStatus.NOT_FOUND || code === grpcStatus.FAILED_PRECONDITION)
+    ) {
+      let runtimeStatus: OrchestrationRuntimeStatus | undefined;
+      try {
+        runtimeStatus = (await this.getStatus(instanceId)).runtimeStatus;
+      } catch {
+        runtimeStatus = undefined;
+      }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 1f4003a. `_mapControlPlaneError` now resolves the instance's runtime state by querying `getOrchestrationState(instanceId)` **directly** instead of routing through `getStatus` inside a broad `catch`:

- resolves a state → map its `runtimeStatus` (terminal → no-op per v3 410; non-terminal → wrong-state handling),
- resolves `undefined` → the genuine "missing" sentinel → `No instance with ID '<id>' found.`,
- **the lookup itself rejects (transient, e.g. an unavailable channel) → the ORIGINAL control-plane error is rethrown**, so a real gRPC/channel failure is never masked as not-found.

Added two unit tests locking this contract (terminate + raiseEvent: control-plane error + `getOrchestrationState` rejecting ⇒ the original error is resurfaced, not converted to not-found). Good catch — thanks.

packages/azure-functions-durable/src/index.ts
YunchuWang: Documented in `ea5cd02` (README migration notes). Since the package CHANGELOG intentionally stays **TBD** until release, the README is now the migration source of truth.

Added a bullet: v3 `context.df.lock(...)` / `context.df.isLocked()` and the `DurableLock` / `LockState` / `LockingRulesViolationError` exports are removed → acquire locks with the core `context.entities.lockEntities(...entityIds)` (returns a `LockHandle` — call `release()`, ideally in a `finally`) and query with `context.entities.isInCriticalSection()`. Restoring the v3 `df.lock` / `isLocked` surface is tracked in #317.

packages/azure-functions-durable/CHANGELOG.md
YunchuWang: Fixed in `ea5cd02` — you were right, `DurableError` / `AggregatedError` / `TokenSource` were never v3 top-level exports, so I removed them from the README's removed-exports bullet.

The README is now the migration source of truth: the package CHANGELOG stays **TBD** until release, so I also dropped the broken `See CHANGELOG.md for the full list` link that pointed at the empty placeholder. The bullet now lists only genuinely-removed surfaces — `DummyOrchestrationContext` / `DummyEntityContext`, `ManagedIdentityTokenSource`, and the entity-lock types — and notes `TaskFailedError` is re-exported from core with aggregate failures surfacing as native `AggregateError`.

packages/azure-functions-durable/src/orchestration-context.ts
YunchuWang: Documented in `ea5cd02` (README migration notes). Added a bullet: `context.df.callHttp(...)` now throws — v3 ran durable HTTP as a host-managed activity and the consolidated gRPC backend has no equivalent primitive, so implement an HTTP activity in your app and call it from the orchestrator. Restoring `callHttp` as a worker-side durable activity is tracked in #318.

packages/azure-functions-durable/src/get-client.ts
```diff
@@ -0,0 +1,71 @@
...
+    const originalStop = created.stop.bind(created);
+    created.stop = async (): Promise<void> => {
+      clientCache.delete(cacheKey);
+      await originalStop();
+    };
+    client = created;
+    clientCache.set(cacheKey, client);
+  }
+  return client;
+}
```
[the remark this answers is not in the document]
YunchuWang: Fixed in `e38c705`. `getClient` now validates/coerces the binding via `asClientInput(...)` **before** deriving the cache key, and a new `deriveCacheKey` helper serializes object configs inside a `try/catch` — returning `undefined` (→ uncached client, no throw) when the config is non-serializable. A missing/`undefined` value now throws the friendly *"not a valid durable client input"* error instead of producing a degenerate key or a raw `JSON.stringify` TypeError.

Cache keys are byte-identical to before for all valid string/object inputs, so caching + evict-on-stop are unchanged. Added tests: missing value → friendly error; a valid object config carrying a `BigInt` field → no throw, and two calls return different (uncached) instances.

packages/durabletask-js/src/worker/task-hub-grpc-worker.ts
```diff
@@ -313,6 +313,69 @@ export class TaskHubGrpcWorker {
...
+      if (stub.abandoned) {
+        // Versioning resolved this work item to the abandon path. Abandon has no meaning on the
+        // single-work-item host path (there is no work-item queue to hand the item back to), so
+        // surface a distinct, actionable error instead of the generic "no response" one. Construct
+        // the worker without versioning when using processOrchestratorRequest.
+        throw new Error(
+          "Orchestrator work item was abandoned (a version mismatch resolved to the abandon path). " +
+            "Abandon is not supported on the single-work-item processOrchestratorRequest path; " +
+            "construct the worker without versioning for this host integration.",
+        );
```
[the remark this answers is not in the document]
YunchuWang: Fixed in `e38c705`. Reworded the abandon-path error (and its comment) to make clear only `VersionFailureStrategy.Reject` (the default abandon strategy) is unsupported on the single-work-item `processOrchestratorRequest` path — `VersionFailureStrategy.Fail` (fails the orchestration in-process and returns a response) or disabling versioning both work. Users are now pointed at setting `failureStrategy` rather than removing versioning wholesale.

andystaples (approved):

PR #284, opened 2026-07-07
refactor: remove dead _failedTasks field and unused async-pageable.ts

### LINKED ISSUE #217: [copilot-finds] Improve: Dead _failedTasks field in CompositeTask and dead async-pageable.ts file
## Problem

Two instances of dead code exist in the core SDK:

### 1. `_failedTasks` field in `CompositeTask` (`src/task/composite-task.ts`, lines 12, 19)

The `_failedTasks` field is declared and initialized to `0` in the constructor, but **never incremented or read** by any code in the entire codebase. Neither `WhenAllTask.onChildCompleted()` nor `WhenAnyTask.onChildCompleted()` ever modifies this counter. The field misleadingly suggests that failed tasks are tracked separately from `_completedTasks`, but they are not.

Additionally, a stale comment in `WhenAllTask` (line 14) references `_failedTasks` as something that should not be re-initialized, but since the field is unused, the comment is misleading.

### 2. Dead `src/utils/async-pageable.ts` file (91 lines)

This file contains a complete `AsyncPageable` class with `Page` interface and `PageFetcher` type that is **never imported by any file** in the entire codebase. The actual pagination implementation lives in `src/orchestration/page.ts` (which is exported from `index.ts` and used by `client.ts`). The dead file creates confusion for developers navigating the codebase.

## Root Cause

The `_failedTasks` field was likely planned as a counter for failed child tasks but was never wired into the `WhenAllTask` or `WhenAnyTask` completion logic. The `async-pageable.ts` file appears to be an earlier implementation that was superseded by `page.ts` but never cleaned up.

## Proposed Fix

1. Remove the `_failedTasks` field from `CompositeTask`
2. Update the stale comment in `WhenAllTask` that references `_failedTasks`
3. Delete the dead `async-pageable.ts` file
4. Add unit tests that explicitly verify the `_completedTasks` counter correctly counts both successful and failed child completions (documenting the intended behavior)

## Impact

Low severity. This is dead code cleanup that reduces maintenance burden and prevents confusion. No behavioral change.

YunchuWang:
## Summary

Dead-code cleanup in the core SDK. **No behavior change** — build, lint, and the full core test suite (66 suites / 1165 tests) all pass after merging `main`.

### 1. Removed the `_failedTasks` field from `CompositeTask`
`_failedTasks` was declared and initialized to `0` in the constructor but **never incremented or read** anywhere in the repository. The only occurrences were the declaration, the constructor init, and a stale comment:

```
packages/durabletask-js/src/task/composite-task.ts:12:  _failedTasks: number;
packages/durabletask-js/src/task/composite-task.ts:19:    this._failedTasks = 0;
packages/durabletask-js/src/task/when-all-task.ts:14:  // ...references _failedTasks
```

Neither `WhenAllTask.onChildCompleted()` nor `WhenAnyTask.onChildCompleted()` ever touched it. The field was **misleading**: it implied failed children were tracked separately, when in reality `CompositeTask` counts *every* settled child — successful **and** failed — in the single `_completedTasks` counter. A separate failure counter was never needed.

This still holds after #301 changed `WhenAllTask` from fail-fast to wait-all: the wait-all path derives failures by filtering `_tasks` for `isFailed` at completion time, so it has no use for a running failure count either.

### 2. Fixed the stale comment in `WhenAllTask`
The constructor comment warned against re-initializing `_completedTasks` **or `_failedTasks`**. With `_failedTasks` gone, the comment now references only `_completedTasks` so it stays accurate.

### 3. Deleted the dead `src/utils/async-pageable.ts` (91 lines)
This file contained an `AsyncPageable` class / `Page` interface / `PageFetcher` type that was **never imported by any file**. The real pagination implementation lives in `src/orchestration/page.ts`, which is exported from `index.ts` and consumed by `client.ts` and the export-history client. Verified before deleting:
- No file references it by path (`grep "async-pageable"` → only the file itself).
- Every `AsyncPageable` / `createAsyncPageable` import resolves to `orchestration/page.ts` (note: `async-pageable.ts` didn't even export `createAsyncPageable`).
- It was not re-exported from any barrel (there is no `utils/index.ts`).

The core build stays green after deletion, confirming nothing depended on it.

### 4. Added tests documenting the real behavior
To capture the intent that `_failedTasks` misleadingly implied was separate:
- **`when-all-task.spec.ts`** — verifies `completedTasks` reaches `2` for one successful and one failed child (single counter, no separate failure count), and that the WhenAll only settles once every child is terminal.
- **`when-any-task.spec.ts`** — verifies `WhenAnyTask` settles on the first completed child regardless of success/failure and does not track failures separately (a later successful child does not override an already-settled failed child).

## Merge with `main`

`main` moved `WhenAllTask` from fail-fast to wait-all with `AggregateError` (#301), which conflicted with the test added here. Resolution: **all of `main`'s tests are kept verbatim**, and the `_failedTasks` documentation test was rewritten for wait-all semantics. Only the two source-file edits and the two added tests remain as this PR's net diff against `main`.

## Review feedback

- @halspang: the third child task in the new WhenAll test was unused — removed, the test now uses only the two children it actually exercises.

## Verification
- `npm run build:core` — clean compile (confirms deleting `async-pageable.ts` broke nothing).
- `npx eslint` on all changed files — exit 0.
- `npx jest --runInBand` (full core suite) — **66 suites, 1165 tests, all pass**.

Fixes #217


packages/durabletask-js/test/when-all-task.spec.ts
```diff
@@ -126,4 +126,25 @@ describe("WhenAllTask", () => {
...
   });
+
+  // Issue #217: failed child completions are counted in the same _completedTasks
+  // counter as successful ones. There is no separate failed-task counter, so a
+  // failing child increments completedTasks just like a successful one before
+  // fail-fast completion kicks in.
+  it("should count both successful and failed children in completedTasks", () => {
+    const child1 = new CompletableTask<number>();
+    const child2 = new CompletableTask<number>();
+    const child3 = new CompletableTask<number>();
```
halspang, line 174: Nit: This is unused aside from the `WhenAllTask`?

kaibocai (dismissed):

packages/durabletask-js/test/when-all-task.spec.ts
YunchuWang: Good catch - `child3` was only there to keep the WhenAll pending under the old fail-fast behavior, and it was never asserted on. Dropped it: the test now uses just the two children it actually exercises (one success + one failure), which is the minimum needed to show both increment the same `_completedTasks` counter. Also had to rework this test anyway, since #301 landed on main and changed WhenAllTask from fail-fast to wait-all.

kaibocai (approved):

PR #300, opened 2026-07-14
Fix rewind deadlock when Task.all had multiple failed siblings (#299)

### LINKED ISSUE #299: Rewind of nested orchestration deadlocks when a Task.all had multiple failed activities (only one failed sibling re-dispatched on replay)
## Summary

Rewind support (added in #296) does not correctly complete a **nested** orchestration when a `Task.all` (fan-out) contained **more than one failed activity**. After a server-side rewind (Azure Storage / DurableTask.Core backend), the worker re-dispatches only **one** of the previously-failed sibling activities and leaves the other as a permanently-unresolved pending task. The child sub-orchestration's `Task.all` never resolves, so **no `SubOrchestrationInstanceCompleted` is emitted after the rewind**, the parent and root orchestrations never complete, and the instance hangs until the caller times out.

This blocks **rewind parity for Azure Functions durable-functions v4** (the consolidation onto durabletask-js): the extension E2E test `RewindOrchestratorTests.RewindFailedOrchestration_ShouldSucceed` **passes on the classic v3 SDK** but **times out on durabletask-js** for the identical app + backend.

## Impact

- Any rewind of a failed orchestration whose failure fanned out through a `Task.all` with **2+ failed activities in the same sub-orchestration** will hang instead of completing.
- Single-failure and single-level cases appear unaffected; the defect is specific to **multiple failed siblings inside one `Task.all`**, in a **sub-orchestration** whose completion must propagate back up.

## Scope / what this is NOT

- Reproduced on the **Azure Storage backend**, where rewind is computed **server-side** by DurableTask.Core. The worker receives an ordinary replay whose history has been rewritten by the backend — it does **not** go through the `EXECUTIONREWOUND` path. `worker/rewind.ts` `buildRewindResult` is **not** exercised here (0 `EXECUTIONREWOUND` events observed), so this is **separate from #296's `buildRewindResult`**.
- **Not** the entity-replay corruption fixed in #294 — 0 `sendEvent ... trying to call a different method` occurrences in the failing run.
- The host performed the rewind correctly (`RewindInstance` returns HTTP 200, `State: Rewound`), and re-dispatched the replay. The stall is **worker-side**: the executor returns `0 actions` while still awaiting a task.

## Reproduction

App: the `RewindOrchestration` sample used by the extension E2E suite (checked into `Azure/azure-functions-durable-extension` at `test/e2e/Apps/BasicNode/src/functions/RewindOrchestration.ts`). Shape:

```
RewindParentOrchestration
  Task.all([ SucceedSub, FailParentSub_1, FailParentSub_2, SucceedSub ])
    FailParentSub -> SucceedActivity -> (callEntity) -> FailChildSub
      FailChildSub:
        Task.all([ SucceedActivity, FailActivity_1, FailActivity_2, (callEntity) ])
```

`FailActivity` throws on the first `numFailures` invocations, then succeeds. Both `FailActivity_1` and `FailActivity_2` fail on the first attempt, so the failure propagates `FailChildSub -> FailParentSub -> root`, leaving the instance `Failed`.

Steps:
1. Start the orchestration (`input=fail`, `numFailures=1`), wait for `Failed`.
2. Call rewind on the instance.
3. Expected: retried activities succeed and the orchestration reaches `Completed`.
4. Actual: the orchestration never reaches `Completed`; the caller times out.

### A/B confirmation (single variable = the JS SDK)

Same app source (verified byte-identical via SHA-256 of every source file), same Azure Storage backend, only the durable-functions/durabletask-js package swapped:

| Arm | SDK | Result |
| --- | --- | --- |
| A (control) | published classic v3 SDK | `RewindFailedOrchestration_ShouldSucceed(1)` and `(2)` **PASS** |
| B (treatment) | durabletask-js consolidated worker (this repo `main`, incl. #293/#294/#296) | both **FAIL** — `Orchestration did not reach Completed status within 30 seconds` |

The v3 SDK completing the identical rewritten history confirms the scenario is completable on Azure Storage and that this is a durabletask-js worker defect, not a backend or sample issue. The extension's own CI corroborates the baseline: the `azurestorage-*(node)` legs (which run this test — it is only `DTS`/`Java`/`Python`/`PowerShell`-skipped, not `Node`/`AzureStorage`-skipped) are green on the published SDK.

## Root cause (worker replay)

Post-rewind, the deepest child `FailChildSubOrchestration` replays a rewritten history that still contains a `TASKSCHEDULED` event for one of the two failed sibling activities, but with its terminal completion event (`TaskFailed`/`TaskCompleted`) stripped by the rewind. Observed in the worker trace (single child instance):

```
Rebuilding local state with 16 history events
... TASKSCHEDULED (6) ... TASKSCHEDULED (6) ...
Waiting for 2 task(s) and 0 event(s)   ->   Returning 1 action(s)     # only ONE failed sibling re-dispatched
FailActivity started -> completed "Success!"                          # that one retry runs and succeeds
Waiting for 1 task(s) and 0 event(s)   ->   Returning 0 action(s)     # DEADLOCK: second sibling orphaned
```

- The failed sibling whose `TaskScheduled` was **removed** by the rewind is correctly re-generated as a new `scheduleTask` action on replay → runs → succeeds.
- The failed sibling whose `TaskScheduled` was **retained** (completion stripped) is mis-handled: `handleTaskScheduled` deletes its entry from `_pendingActions` (assuming a completion will arrive later), but no completion ever arrives after a rewind. The corresponding `_pendingTasks` entry is never resolved and is never re-dispatched → the `Task.all` blocks forever.
- Because the child never resolves, **0 `SubOrchestrationInstanceCompleted`** are emitted after the rewind; the parent and root `Task.all` never resolve; no root `ExecutionCompleted` → timeout.

Most-likely fix direction: during replay reconciliation, a task that is `scheduled-but-not-completed` because its terminal event was rewound away must be **re-dispatched** (its `scheduleTask` action retained), rather than treated as "already in flight" and dropped. This needs to cover **all** such orphaned siblings in a `Task.all`, not just the ones whose `TaskScheduled` the backend also removed. Verifying the `_pendingTasks` / `_pendingActions` bookkeeping in `orchestration-executor.ts` (`handleTaskScheduled`, `getActions`) for the rewound-history shape is the place to start.

### Minor secondary observation
During replay the worker logs `Unknown history event type: GENERICEVENT (18), skipping...` and `... ORCHESTRATORCOMPLETED (15), skipping...`. These appear benign (the run's failure is fully explained by the orphaned-task deadlock above), but the skipped `GENERICEVENT`/`ORCHESTRATORCOMPLETED` handling in the rewound-history path may be worth a confirming glance.

## Suggested test

Add an in-memory executor/e2e case: a sub-orchestration whose `Task.all` contains **two** activities that both fail on first attempt; drive it to `Failed`, rewind, and assert the sub-orchestration and its parent reach `Completed` with each failed activity invoked exactly `numFailures + 1` times. (This mirrors `RewindOrchestratorTests.RewindFailedOrchestration_ShouldSucceed` in the extension.)

YunchuWang:
## Summary

Fixes a rewind deadlock where a **nested** orchestration whose `Task.all` fan-out contained **more than one failed activity** never completes after a server-side (Azure Storage / DurableTask.Core) rewind. The worker re-dispatched only **one** previously-failed sibling and left the other as a permanently-unresolved pending task, so the child's `Task.all` never resolved, no `SubOrchestrationInstanceCompleted` was emitted, and the parent/root hung until the 30s timeout.

`Closes #299`

## Root cause (verified)

On the Azure Storage backend, rewind is computed **server-side** by DurableTask.Core — this is a plain replay of a backend-rewritten history, **not** the `EXECUTIONREWOUND` / `buildRewindResult` path from #296 (0 `ExecutionRewound` events observed).

The rewind (`DurableTask.AzureStorage` `RewindHistoryAsync` / `DurableTask.Core` `ProcessRewindOrchestrationDecision`) removes/morphs a failed activity's `TaskScheduled` **only when a `TaskFailed` event exists** for it. Because `Task.all` fails fast, the child completes-failed after the **first** sibling failure, so later siblings' `TaskFailed` events are never committed (late completion for a terminal instance is dropped). After rewind:

- **Sibling A** (`TaskFailed` present): its `TaskScheduled` is morphed away → its pending action survives replay → re-dispatched cleanly. ✅
- **Sibling B** (`TaskFailed` missing): its `TaskScheduled` is left **bare** (no terminal) → `handleTaskScheduled` deletes its pending action assuming a completion will arrive, but after a rewind none ever does → orphaned pending task → `Task.all` deadlocks. ❌

Observed worker trace (single child instance): `Waiting for 2 task(s) … Returning 1 action(s)` (one sibling re-dispatched, runs, succeeds) then `Waiting for 1 task(s) … Returning 0 action(s)` = deadlock.

## Fix (worker executor)

1. **Detect a rewind revival from the NEW events.** Azure Storage wakes a rewound instance with a `GenericEvent` (`RewindTaskOrchestrationAsync` sends `new GenericEvent(-1, reason)`); the DTS jump-start path uses `ExecutionRewound`. Gating on new events scopes reconciliation to the revival work item, so ordinary post-rewind replays (whose new events are normal completions) never re-dispatch genuinely in-flight tasks.
2. **Re-dispatch orphaned activities.** During replay, archive each consumed `scheduleTask` action. At the end of a revival replay, for every still-pending task whose `scheduleTask` action was consumed by a bare `TaskScheduled` but has no live pending action, re-add the archived action. **General for N failed siblings.** Sub-orchestrations are intentionally excluded (their retained `SubOrchestrationInstanceCreated` is revived separately by the backend).
3. **Make duplicate `TaskScheduled` idempotent.** `DurableTask.Core` `ProcessScheduleTaskDecision` always appends a fresh `TaskScheduled` per action (no dedup), so re-dispatching the bare orphan produces a **second** `TaskScheduled` with the same id. A repeat for an already-handled id is now treated as idempotent instead of raising `NonDeterminismError`.

57 lines of source across 2 files (`orchestration-executor.ts`, `runtime-orchestration-context.ts`).

## Test evidence

New deterministic executor-level regression test `test/rewind-multi-sibling-deadlock.spec.ts` drives the executor through episodes modelling the classic Azure Storage backend (append-only `TaskScheduled`, revival `GenericEvent` trigger):

- **re-dispatches ALL previously-failed `Task.all` siblings after a rewind and completes** — red without the fix (deadlock), green with it.
- **re-dispatches N (=3) failed siblings** — generality check.
- **does NOT re-dispatch an in-flight bare `TaskScheduled` during ordinary (non-rewind) replay** — no-regression guard (passes on `main` and with the fix).

Red/green proof (identical test code):

```
# source reverted (main behavior):
× re-dispatches ALL previously-failed Task.all siblings ...   (deadlocked: expected false, got true)
× re-dispatches N (=3) failed siblings ...
√ does NOT re-dispatch an in-flight bare TaskScheduled ...
Tests: 2 failed, 1 passed

# with fix:
√ re-dispatches ALL previously-failed Task.all siblings ...
√ re-dispatches N (=3) failed siblings ...
√ does NOT re-dispatch an in-flight bare TaskScheduled ...
Tests: 3 passed
```

## Sister-SDK parity (corroboration)

The same defect exists in **durabletask-python** (independently verified against `microsoft/durabletask-python@e7b5f15`), which confirms this is a **core, cross-SDK replay-contract issue** rather than a JS-specific quirk:

- Its `taskScheduled` handler (`worker.py` ~L2196-2205) pops the pending action on a bare `TaskScheduled` and **raises a non-determinism error if the action is already gone** — `action = ctx._pending_actions.pop(task_id, None); if not action: raise _get_non_determinism_error(...)` — while only *reading* (`.get`) the pending task. So a rewound-away sibling leaves an orphaned pending task, exactly as in JS.
- `get_actions()` (`worker.py` ~L1549) returns `list(self._pending_actions.values())`, so a popped orphan is **never re-dispatched** → identical `Task.all` deadlock after a multi-failed-sibling rewind.
- Python's rewind support (`_build_rewind_result`) only short-circuits the **DTS jump-start** path (an `executionRewound` event in the new events). A **classic Azure Storage server-side rewind** (no `executionRewound`) falls through to normal replay — the same path that deadlocks here — and there is no `_consumed` / rewind-revival re-dispatch and no idempotent duplicate-`TaskScheduled` guard. The durabletask-python functions-support PR microsoft/durabletask-python#155 only added a `genericEvent` *ignore*, not re-dispatch, so the deadlock remains **unfixed** there.

This independently corroborates two things about this PR:

1. **The root cause is at the DurableTask replay-contract level**, not a JS bug.
2. **The idempotent handled-id guard is load-bearing.** Python's `if not action: raise NonDeterminism` on a repeat `TaskScheduled` is exactly what a naive re-dispatch would trip — because re-dispatching the orphan makes the backend append a *second* `TaskScheduled` for the same id. Re-dispatch **without** the idempotency guard would convert the deadlock into a `NonDeterminismError`, which is why this PR pairs re-dispatch with idempotent duplicate-`TaskScheduled` handling.

A durabletask-python parity fix is therefore warranted and should be tracked separately.

## E2E status on real Azure Storage (verified)

Run on a live Azure Storage backend (fix logic ported onto the #282 gRPC-worker core; fix anchors byte-identical; `callEntities=false`, single target framework to force one deterministic instance):

- The **NonDeterminism / duplicate-`TaskScheduled` assumption is CLOSED** — zero `NonDeterminismError` across `numFailures` 1 and 2.
- On the path DurableTask.Core actually rewinds, the **signaled** child logs `Waiting for 2 → Returning 2` (both failed activity siblings re-dispatched) and completes, whereas the **un-fixed baseline** logs `Returning 1` and deadlocks — the decisive proof this fix is real and necessary.

This fix is **necessary but not sufficient** to turn `RewindOrchestratorTests.RewindFailedOrchestration_ShouldSucceed(1|2)` green, because that E2E **also** exercises a distinct, deeper gap: the **root** `Task.all` fails fast on the first sub-orchestration failure, so the second failing sub-orch's failure is dropped (late completion for a now-terminal parent). Server-side rewind therefore signals **only** the committed-failed subtree; the other failing subtree is re-driven with **no** rewind signal, so its grandchild hits the identical #299 orphan that the (correctly gate-guarded) reconcile cannot repair without a signal. Which subtree is dropped is a **race**. That unsignaled-sibling gap is tracked separately in #301.

## Gates

- `npm run build` — clean (all workspaces)
- `npm run lint` — clean
- `npm run test:unit` — **1283 passed, 0 failed** (1122 core incl. the new spec and the #296 `rewind-e2e` suite, 106 azuremanaged, 55 export-history)

## Notes / honesty

- Root cause and fix were verified against the real `DurableTask.Core` / `DurableTask.AzureStorage` rewind source, the failing-run worker log, and the issue #299 A/B evidence.
- The shipped correctness proof is the **executor-level** regression test, which faithfully models the append-only backend and the duplicate-`TaskScheduled` behavior the idempotent guard handles. The full extension E2E was **additionally** run live on Azure Storage (see **E2E status** above): on the path DTFx actually rewinds it confirms the #299 fix — the signaled grandchild logs `Waiting for 2 → Returning 2` (both failed siblings re-dispatched) and completes, versus the NOFIX baseline's `Returning 1` deadlock — with **zero `NonDeterminismError`** across every run, so the idempotent handled-id guard holds on the real backend. **However**, clean single-target-framework runs of `RewindFailedOrchestration_ShouldSucceed(1|2)` still **fail** at `RewindOrchestratorTests.cs:50` for a distinct, deeper reason — an unsignaled re-driven sibling subtree, tracked in #301 — and an earlier apparent pass was a **multi-TFM (`net8.0;net10.0`) concurrency artifact**, not a reliable green. #300 stays correctly scoped to the #299 signaled-path fix.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/packages/durabletask-js/src/worker/orchestration-executor.ts
+++ b/packages/durabletask-js/src/worker/orchestration-executor.ts
@@ -106,6 +106,24 @@ export class OrchestrationExecutor {
       ctx.setFailed(e instanceof Error ? e : new Error(String(e)));
     }
 
+    // Issue #299: On a server-side (Azure Storage / DurableTask.Core) rewind, the backend strips
+    // the terminal event (TaskFailed/TaskCompleted) of previously-failed activities. When a
+    // Task.all fan-out had more than one failed sibling, the child completes-failed after the
+    // FIRST failure (fail-fast), so a later sibling's TaskFailed is never committed and its
+    // TaskScheduled is left BARE. Ordinary replay deletes that activity's pending action
+    // (handleTaskScheduled) assuming a completion will arrive, but after a rewind none ever will
+    // -> the task is orphaned and the Task.all deadlocks. When this replay is a rewind revival,
+    // re-dispatch every such orphaned activity so ALL previously-failed siblings rerun.
+    if (!ctx._isComplete && this.isRewindRevival(newEvents)) {
+      for (const key of Object.keys(ctx._pendingTasks)) {
+        const taskId = Number(key);
+        const consumedAction = ctx._consumedActivityActions[taskId];
+        if (consumedAction && !ctx._pendingActions[taskId]) {
+          ctx._pendingActions[taskId] = consumedAction;
+        }
+      }
+    }
+
     if (!ctx._isComplete) {
       const taskCount = Object.keys(ctx._pendingTasks).length;
       const eventCount = Object.keys(ctx._pendingEvents).length;
@@ -130,6 +148,19 @@ export class OrchestrationExecutor {
       actions,
       customStatus: ctx.getCustomStatus(),
     };
+  }
+
+  /**
+   * Detects whether this replay is a rewind revival, i.e. the current work item was triggered by a
+   * rewind rather than ordinary progress. On the Azure Storage / DurableTask.Core backend the
+   * revived instance is woken with a `GenericEvent` message (see
+   * DurableTask.AzureStorage `RewindTaskOrchestrationAsync`, which sends `new GenericEvent(-1, ...)`);
+   * the DTS "jump-start" path instead delivers an `ExecutionRewound` event. Gating on the NEW
+   * events keeps reconciliation scoped to the revival work item, so ordinary post-rewind replays
+   * (whose new events are normal completions) never re-dispatch genuinely in-flight tasks.
+   */
+  private isRewindRevival(newEvents: pb.HistoryEvent[]): boolean {
+    return newEvents.some((e) => e.hasExecutionrewound() || e.hasGenericevent());
   }
 
   private async processEvent(ctx: RuntimeOrchestrationContext, event: pb.HistoryEvent): Promise<void> {
@@ -355,6 +386,17 @@ export class OrchestrationExecutor {
   private async handleTaskScheduled(ctx: RuntimeOrchestrationContext, event: pb.HistoryEvent): Promise<void> {
     // This history event confirms that the activity execution was successfully scheduled. Remove the taskscheduled event from the pending action list so we don't schedule it again.
     const taskId = event.getEventid();
+
+    // Issue #299: A rewind revival re-dispatches an activity whose terminal event was stripped.
+    // DurableTask.Core appends a fresh TaskScheduled per scheduleTask action with no dedup, so the
+    // re-dispatched activity yields a SECOND TaskScheduled with the same id as the retained bare
+    // one. Duplicate TaskScheduled ids cannot occur in normal deterministic replay, so treat a
+    // repeat as idempotent (the pending action was already consumed on the first occurrence)
+    // instead of raising a non-determinism error.
+    if (ctx._handledTaskScheduledIds.has(taskId)) {
+      return;
+    }
+
     const action = ctx._pendingActions[taskId];
     delete ctx._pendingActions[taskId];
 
@@ -373,6 +415,12 @@ export class OrchestrationExecutor {
         action.getScheduletask()?.getName(),
       );
     }
+
+    // Remember the consumed scheduleTask action keyed by its id. If this replay turns out to be a
+    // rewind revival and the activity's terminal event was stripped (leaving this bare
+    // TaskScheduled with no completion), end-of-replay reconciliation re-dispatches it.
+    ctx._handledTaskScheduledIds.add(taskId);
+    ctx._consumedActivityActions[taskId] = action;
   }
 
   private async handleTaskCompleted(ctx: RuntimeOrchestrationContext, event: pb.HistoryEvent): Promise<void> {
--- a/packages/durabletask-js/src/worker/runtime-orchestration-context.ts
+++ b/packages/durabletask-js/src/worker/runtime-orchestration-context.ts
@@ -51,6 +51,13 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
   _saveEvents: boolean;
   _customStatus?: string;
   _entityFeature: RuntimeOrchestrationEntityFeature;
+  // Issue #299: bookkeeping used to recover from a server-side rewind. `_consumedActivityActions`
+  // remembers the scheduleTask action behind every TaskScheduled event seen during replay so an
+  // orphaned activity (its terminal event stripped by the rewind) can be re-dispatched.
+  // `_handledTaskScheduledIds` makes a duplicate TaskScheduled for an already-handled id idempotent
+  // (a post-rewind re-dispatch appends a fresh TaskScheduled with the same id -- no backend dedup).
+  _consumedActivityActions: Record<number, pb.OrchestratorAction>;
+  _handledTaskScheduledIds: Set<number>;
 
   constructor(instanceId: string) {
     super();
@@ -74,6 +81,8 @@ export class RuntimeOrchestrationContext extends OrchestrationContext {
     this._saveEvents = false;
     this._customStatus = undefined;
     this._entityFeature = new RuntimeOrchestrationEntityFeature(this);
+    this._consumedActivityActions = {};
+    this._handledTaskScheduledIds = new Set<number>();
   }
 
   get instanceId(): string {
--- a/packages/durabletask-js/test/rewind-multi-sibling-deadlock.spec.ts
+++ b/packages/durabletask-js/test/rewind-multi-sibling-deadlock.spec.ts
@@ -1,0 +1,287 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// Regression tests for issue #299:
+// https://github.com/microsoft/durabletask-js/issues/299
+//
+// A nested orchestration whose `Task.all` fan-out had MORE THAN ONE failed activity fails to
+// complete after an Azure Storage server-side rewind. The backend rewrites the failed
+// orchestration's history; for each failed activity it removes/morphs the corresponding
+// `TaskScheduled` ONLY when a `TaskFailed` event exists for it (DurableTask.Core
+// `ProcessRewindOrchestrationDecision`; DurableTask.AzureStorage `RewindHistoryAsync`).
+//
+// When `Task.all` fails fast, the child orchestration completes-failed after the FIRST sibling
+// failure and the SECOND sibling's `TaskFailed` is never committed. After the rewind that
+// sibling is left with a BARE `TaskScheduled` (no terminal event). On replay the worker deletes
+// its pending action (assuming a completion will arrive) but none ever does -> the child's
+// `Task.all` never resolves -> deadlock -> 30s timeout.
+//
+// These tests reproduce the exact post-rewind history at the executor level and drive it through
+// a small harness that models the classic Azure Storage backend (which ALWAYS appends a fresh
+// `TaskScheduled` per scheduleTask action -- DurableTask.Core `ProcessScheduleTaskDecision`,
+// no dedup).
+
+import { StringValue } from "google-protobuf/google/protobuf/wrappers_pb";
+import { Timestamp } from "google-protobuf/google/protobuf/timestamp_pb";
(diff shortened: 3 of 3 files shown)


YunchuWang:
Closing in favor of a **root-cause fix** (tracked in #301).

**Why:** Analysis established that this PR's worker-side *archive + revive* is a band-aid that (1) only repairs the orphan on the rewind-**signaled** path and provably cannot fix the unsignaled re-driven sub-orchestration case (#301), and (2) becomes dead code once the real cause is removed.

**The real cause** is `WhenAllTask` **fail-fast**: `onChildCompleted` completes the `whenAll` on the *first* child failure, so a multi-failed-sibling `whenAll` lets the orchestration go terminal before the later siblings' `TaskFailed` events are committed. Those siblings keep a bare `TaskScheduled` with no completion, and on rewind the worker pops the pending action but no completion ever arrives -> the `whenAll` waits forever -> deadlock.

**Cross-SDK evidence:** .NET (`System.Threading.Tasks.Task.WhenAll`) and Java (`CompletableFuture.allOf`) both **wait for all** children and **aggregate** exceptions — they never drop a sibling failure and are structurally immune to this bug. durabletask-js and durabletask-python diverged by adding fail-fast.

**The fix** is to align durabletask-js `WhenAll` to **wait-all** semantics (complete only after every child reaches a terminal state, even on failure). That removes the orphan at the source — at every nesting level, signaled or not — and makes this revive machinery unnecessary. A replacement PR implementing the wait-all fix (JS-only) is being opened and will reference #301. durabletask-python shares the same fail-fast root cause and will be addressed separately.

YunchuWang:
Replacement PR is now open: #302 (`Fixes #301`) — the wait-all root-cause fix described above. It changes one source file (`when-all-task.ts`) to complete `WhenAll` only after all children are terminal, surfacing the first failure without aggregation, and updates the existing fail-fast tests plus adds a two-batch multi-failed-sibling regression test for #301. No archive/revive machinery.

PR #307, opened 2026-07-17
fix: add missing CANCELED orchestration status

### LINKED ISSUE #203: [copilot-finds] Bug: Missing CANCELED member in OrchestrationStatus enum causes client crashes
## Problem

The client-facing `OrchestrationStatus` enum in `packages/durabletask-js/src/orchestration/enum/orchestration-status.enum.ts` (line 27-35) is missing the `CANCELED` member, even though the protobuf definition includes `ORCHESTRATION_STATUS_CANCELED = 4`.

This causes `fromProtobuf()` to throw `"Unknown protobuf OrchestrationStatus value: 4"` when encountering any orchestration with the CANCELED status. Since `fromProtobuf()` is called by `newOrchestrationState()`, this crashes every client operation that reads orchestration state:
- `getOrchestration()` 
- `waitForOrchestrationCompletion()`
- `getAllInstances()` / `listInstances()`

Additionally:
- `InMemoryOrchestrationBackend.toClientStatus()` silently maps CANCELED to RUNNING (incorrect)
- `isTerminalStatus()` in both `InMemoryOrchestrationBackend` and `TestOrchestrationClient` does not include CANCELED, so `waitForCompletion()` never returns for canceled orchestrations in the test backend

## Root Cause

When the `OrchestrationStatus` enum was defined, the `CANCELED` member was omitted. All other proto status values (RUNNING, COMPLETED, CONTINUED_AS_NEW, FAILED, TERMINATED, PENDING, SUSPENDED) have corresponding client enum members, but `ORCHESTRATION_STATUS_CANCELED = 4` does not.

## Proposed Fix

1. Add `CANCELED = pb.OrchestrationStatus.ORCHESTRATION_STATUS_CANCELED` to the `OrchestrationStatus` enum (the existing initialization loop auto-registers it in conversion maps)
2. Add CANCELED case to `InMemoryOrchestrationBackend.toClientStatus()`
3. Add CANCELED to `isTerminalStatus()` in both `InMemoryOrchestrationBackend` and `TestOrchestrationClient`
4. Add comprehensive tests for OrchestrationStatus enum conversions

## Impact

**Severity: Medium-High.** Any orchestration that enters the CANCELED state (via the sidecar, another SDK, or future API) would cause the JS client to crash with an unhandled error when querying or listing orchestrations. This prevents interop with other SDKs that may use the CANCELED status.

YunchuWang:
## Summary

- Expose the protobuf CANCELED status through the public orchestration status enum.
- Treat canceled in-memory orchestrations as terminal and map their client status correctly.
- Add conversion and in-memory backend coverage.

Fixes #203

## Testing

- `npm test -w @microsoft/durabletask-js -- --runTestsByPath test/orchestration-status.spec.ts test/in-memory-backend.spec.ts`
- `npm run build:core`


YunchuWang:
Closing as not an active bug. A cross-SDK/backend check shows nothing currently produces `ORCHESTRATION_STATUS_CANCELED` (value 4): durabletask-go defines the constant and treats it as terminal in `IsComplete()` but never assigns it; .NET (DurableTask.Core / AzureStorage) references it only in comparisons; Python has it only in generated proto; and the durabletask-js worker only ever emits COMPLETED/FAILED/TERMINATED. Because no backend emits status 4, `fromProtobuf(4)` is never reached and the described client crash cannot occur in practice. This is a latent/defensive concern rather than a reproducible failure, so closing for now. The deeper fragility (that `fromProtobuf` throws on any unmapped proto status) can be revisited separately if graceful degradation is desired.

PR #313, opened 2026-07-17
fix: preserve export progress statistics on batch failure

YunchuWang:
## Summary

- Commit the actual scanned and exported counts when an export batch fails.
- Preserve the reported failure details instead of resetting progress to zero.
- Add orchestrator regression coverage for partial batch failures.

Fixes #213

## Testing

- `npm test -w @microsoft/durabletask-js-export-history -- --runTestsByPath test/export-job-orchestrator.spec.ts`
- `npm run build -w @microsoft/durabletask-js-export-history`

### DIFF
--- a/packages/durabletask-js-export-history/src/orchestrators/export-job-orchestrator.ts
+++ b/packages/durabletask-js-export-history/src/orchestrators/export-job-orchestrator.ts
@@ -159,8 +159,8 @@ export const exportJobOrchestrator: TOrchestrator = async function* (
         }
       } else {
         yield* commitCheckpoint(ctx, jobEntityId, {
-          scannedInstances: 0,
-          exportedInstances: 0,
+          scannedInstances: scannedCount,
+          exportedInstances: batchResult.exportedCount,
           failures: batchResult.failures,
         } as CommitCheckpointRequest);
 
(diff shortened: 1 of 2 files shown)


YunchuWang:
Closing after a first-principles cross-SDK review.

**The issue's premise ("copy-paste oversight") is not supported.** The .NET sibling `ExportJobOrchestrator` deliberately uses the same `scannedInstances: 0, exportedInstances: 0, checkpoint: null` on the terminal-failure path, with an explicit comment: *"Batch failed after all retries - commit without checkpoint (don't move cursor), record failures."* export-history exists only in JS + .NET, so .NET is the sole cross-SDK reference and it treats `0/0` as intentional.

**Why `0/0` is defensive, not accidental:** checkpoint counts accumulate by addition, and on failure the cursor is intentionally *not* advanced (`commit-checkpoint-request.ts` documents this as "allowing retry of the same batch"). Crediting real counts on a failed attempt would double-count if that page is ever reprocessed.

**Net:** in the current code failure is terminal (throws → `markAsFailed`; no resume op; `create` resets counters), so this change is harmless today and does make the final report more accurate. But it (a) diverges from .NET's deliberate behavior, breaking the repo's cross-SDK consistency convention, and (b) reintroduces a latent double-count risk if a resume-from-failed feature is ever added. This needs a cross-SDK semantics decision by maintainers rather than a one-sided JS change, so closing as not planned.

PR #341, opened 2026-07-30
feat(testing): implement entity support in the in-memory backend

YunchuWang:
## Problem

Any orchestration that touches an entity is **untestable** against the in-memory backend today. `processAction` has no case for `SENDENTITYMESSAGE`, so it hits the `default` branch and throws.

Proven on unpatched `main` — an orchestration whose only statement is `ctx.entities.signalEntity(...)`:

```
PROBE_RESULT status=3 failure=Unknown orchestrator action type '8' for orchestration '...'.
This likely means the in-memory backend needs to be updated to handle a newly introduced action type.
```

`status=3` is `FAILED`. Not a dropped message — the orchestration dies. Same probe with this PR:

```
PROBE_RESULT status=COMPLETED failure=none
```

## Why implement rather than fail fast

I checked all four sibling SDKs for precedent:

| SDK | In-memory testing backend | Entities in it |
|---|---|---|
| **Python** | `durabletask/testing/in_memory_backend.py` (1814 lines) | ✅ full |
| **.NET** | `InProcessTestHost/Sidecar/InMemoryOrchestrationService.cs` (944 lines) | ❌ 0 references |
| **Java** | none | n/a |
| **Go** | none (sqlite + postgres only) | n/a |
| **JS** | `src/testing/in-memory-backend.ts` (899 lines) | ❌ 0 references |

Python is the reference. Its `_process_action` dispatches `sendEntityMessage` to `_process_send_entity_message_action`, which branches on the same four oneof variants this PR handles. This change follows that design.

Worth noting: Python's dispatch handles exactly six actions and has **no** `terminateOrchestration` — independently matching the JS in-memory backend's own action set.

## What changed

**`in-memory-backend.ts`**
- `EntityState`, `EntityWorkItem`, `PendingLockRequest`; entity map, queue, and in-flight set.
- `processSendEntityMessageAction` handles all four variants. Each echoes its confirmation event into orchestration history keyed by the originating **action ID** — required so replay validation (`validateEntityAction`) matches on the next execution.
- `getNextEntityWorkItem()` drains all queued operations into one batch and marks the entity in-flight so it is not dispatched concurrently.
- `completeEntityTask()` persists state, delivers results, applies side effects, and advances the completion token so stale completions are rejected.
- Result correlation is by **index against the dispatched operations**, which is sound because `TaskEntityShim.executeAsync` pushes exactly one result per operation. `callEntity` responses become `EntityOperationCompleted` / `EntityOperationFailed` on the caller; signals are fire-and-forget and produce none.
- Lock manager: `canGrantLock` / `grantLock` / `tryGrantLock` / `tryGrantPendingLocks`; unlock releases and re-evaluates the pending queue.
- Entity side effects: `sendSignal` and `startNewOrchestration`.
- `reset()` clears entity state; `hasPendingWork()` includes the entity queue.

**`test-worker.ts`** — `addEntity` / `addNamedEntity`, entity polling in the processing loop, and batch execution through the existing `TaskEntityShim`. A call to an **unregistered** entity now fails that operation with an explanatory error instead of hanging the caller forever.

**`test-client.ts`** — `signalEntity()` and `getEntity()`, matching Python's `SignalEntity` / `GetEntity` RPCs and the real `TaskHubGrpcClient` signatures.

### Ordering detail

`grantLock` pushes `EntityLockGranted` into `parent.pendingEvents` while that same instance's actions are being processed. This is safe because `completeOrchestration` clears `pendingEvents` *before* the action loop, not after. The lock test covers it.

## Tests

11 new tests in `test/in-memory-backend-entities.spec.ts`. They assert **observable effects**, not just absence of a crash — entity state actually mutates, and `callEntity` return values reach the orchestration:

- signal from an orchestration mutates entity state
- multiple signals batch into one execution (1+2+3 → 6)
- `callEntity` returns its value to the caller
- state persists across separate batches
- a throwing operation surfaces to the caller as a catchable error
- calling an unregistered entity fails instead of hanging
- `lockEntities` grants, then `release()` unlocks
- client-initiated `signalEntity` reaches the entity
- `getEntity` returns `undefined` for a nonexistent entity
- an entity signalling another entity is delivered
- `reset()` clears entity state

## Validation

| Check | Result |
|---|---|
| `npm run build` | exit 0 |
| `npm run lint` | exit 0 |
| `npx jest` (durabletask-js) | **1187 passed / 1187**, 68/68 suites |
| Revert probe | orchestration `FAILED` without the fix, `COMPLETED` with it |

Baseline is 67 suites / 1176 tests; this adds 1 suite / 11 tests. No regressions.

The `azure-functions-durable` package reports 8 suites failing on `Cannot find module '@azure/functions'`. I confirmed this is **pre-existing** by stashing this branch's changes and re-running that package — identical 8 failed / 2 failed. It is a dependency-install gap in my environment, unrelated to this change.

## Scope

Matching Python, an entity is dispatched whenever it has queued operations; operations are not filtered by critical section while a lock is held. Lock state is tracked and `lockEntities` blocks other lock requests, but a concurrent *signal* to a locked entity is delivered immediately rather than deferred. Full critical-section message filtering is a larger change and is not attempted here.

### DIFF
--- a/packages/durabletask-js/src/testing/in-memory-backend.ts
+++ b/packages/durabletask-js/src/testing/in-memory-backend.ts
@@ -491,7 +491,7 @@ export class InMemoryOrchestrationBackend {
     }
 
     const parent = this.instances.get(parentInstanceId);
-    if (!parent) {
+    if (!parent || parent.executionId !== called.getParentexecutionid()?.getValue()) {
       return;
     }
 
--- a/packages/durabletask-js/src/testing/in-memory-backend.ts
+++ b/packages/durabletask-js/src/testing/in-memory-backend.ts
@@ -50,8 +50,12 @@ export interface EntityState {
   instanceId: string;
   serializedState?: string;
   lastModifiedAt: Date;
-  /** Critical section ID currently holding the lock, if any. */
+  /** Orchestration instance currently holding the lock, if any. */
   lockedBy?: string;
+  /** Execution of the orchestration that owns the current critical section. */
+  lockOwnerExecutionId?: string;
+  /** Critical section ID used to correlate the matching unlock action. */
+  lockCriticalSectionId?: string;
   /** Operations queued but not yet dispatched to a worker. */
   pendingOperations: pb.HistoryEvent[];
   /** Operations handed to a worker and awaiting a batch result. */
@@ -65,8 +69,17 @@ export interface EntityState {
 interface PendingLockRequest {
   criticalSectionId: string;
   parentInstanceId: string;
+  parentExecutionId: string;
   lockSet: string[];
+  sequenceNumber: number;
 }
+
+interface RewindSnapshotEntry {
+  executionId: string;
+  completionToken: number;
+}
+
+type RewindSnapshot = Map<string, RewindSnapshotEntry>;
 
 /**
  * Entity work item that needs to be executed.
@@ -108,11 +121,14 @@ export class InMemoryOrchestrationBackend {
   private readonly entityQueue: string[] = [];
   private readonly entityQueueSet: Set<string> = new Set();
   private readonly entityInFlight: Set<string> = new Set();
+  private readonly entityOperationSequence: WeakMap<pb.HistoryEvent, number> = new WeakMap();
   private pendingLockRequests: PendingLockRequest[] = [];
   private readonly stateWaiters: Map<string, StateWaiter[]> = new Map();
   private readonly pendingTimers: Set<ReturnType<typeof setTimeout>> = new Set();
   private readonly instanceTimers: Map<string, Set<ReturnType<typeof setTimeout>>> = new Map();
+  private readonly rewindSnapshots: Map<string, RewindSnapshot> = new Map();
   private nextCompletionToken: number = 1;
+  private nextEntityMessageSequence: number = 1;
   private readonly maxHistorySize: number;
 
   /**
@@ -295,6 +311,8 @@ export class InMemoryOrchestrationBackend {
       return false;
     }
 
+    this.cleanupLocksForExecution(instance.instanceId, instance.executionId);
+    this.clearRewindSnapshot(instanceId);
     this.instances.delete(instanceId);
     this.orchestrationQueueSet.delete(instanceId);
     const queueIndex = this.orchestrationQueue.indexOf(instanceId);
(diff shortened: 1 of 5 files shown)


packages/durabletask-js/src/testing/in-memory-backend.ts
```diff
@@ -316,6 +377,193 @@ export class InMemoryOrchestrationBackend {
...
+      return;
+    }
+
+    const parentInstanceId = called.getParentinstanceid()?.getValue();
+    if (!parentInstanceId) {
+      return;
+    }
+
+    const parent = this.instances.get(parentInstanceId);
+    if (!parent) {
```
bachuv: Should we check for `parent.executionId === called.getParentexecutionid()?.getValue()` here too? Right now we match the caller by instance ID only, so if that instance was purged and recreated, the result could get delivered to the wrong execution.


```suggestion
    if (!parent || parent.executionId !== called.getParentexecutionid()?.getValue()) {
```

packages/durabletask-js/src/testing/in-memory-backend.ts
```diff
@@ -316,6 +357,176 @@ export class InMemoryOrchestrationBackend {
...
+    entity.dispatchedOperations = [];
+    entity.completionToken = this.nextCompletionToken++;
+    this.entityInFlight.delete(instanceId);
+
+    // Deliver results to callers. Results are index-aligned with the dispatched
+    // operations because the executor pushes exactly one result per operation.
+    const results = result.getResultsList();
+    for (let i = 0; i < dispatched.length; i++) {
+      this.deliverEntityOperationResult(dispatched[i], results[i]);
+    }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 12189aa. `createEntityFailureResult()` now receives the original `EntityWorkItem.entityState` and copies it into every pre-shim failure result, covering both the missing-registration path and caught factory/shim exceptions. The new `should preserve entity state when the entity factory fails` regression seeds the counter at 4, forces factory construction to fail, verifies the caller observes that failure, and then verifies the persisted value is still 4 and readable after the factory recovers.

packages/durabletask-js/src/testing/in-memory-backend.ts
YunchuWang: Fixed in 12189aa. `deliverEntityOperationResult()` now requires the current parent execution ID to exactly match `parentExecutionId` before enqueueing a completion. The new `should not deliver a stale entity result to a recreated orchestration` regression holds an entity call in flight, terminates and purges its caller, recreates the same instance ID with a new execution, releases the stale batch, and verifies the old result is absent from the recreated execution before that orchestration completes normally.

kaibocai (approved):

